PR #1, opened 2021-06-29
Ansible deployment

venkatabhishek:
This PR is for mostly for review of ansible setup; the ultimate location for these files is unknown (in terms of for CI/CD)

Notes:
- `docker.yml` is a development playbook designed to start docker services for postgres, rabbitmq, and minio
- `main.yml` is the main playbook to setup celery + django on gunicorn as systemd services. Assumes that environment variables have been set (i.e. `source dev/export-env.sh`)


ansible/main.yml
```diff
@@ -0,0 +1,110 @@
...
+  - name: checkout code
+    git:
+      repo: "https://github.com/ResonantGeoData/RD-WATCH.git"
+      dest: "{{ app_path }}"
+      force: yes
+
+  - name: Set environment variables
+    set_fact:
+      watch_environment:
+        DJANGO_CONFIGURATION:  "{{ lookup('env', 'DJANGO_CONFIGURATION') }}"
```
brianhelba, line 16:
```suggestion
        DJANGO_CONFIGURATION: AWSProductionConfiguration
```

ansible/main.yml
```diff
@@ -0,0 +1,110 @@
...
+        DJANGO_DATABASE_URL: "{{ lookup('env', 'DJANGO_DATABASE_URL') }}"
+        DJANGO_CELERY_BROKER_URL: "{{ lookup('env', 'DJANGO_CELERY_BROKER_URL') }}"
+        DJANGO_MINIO_STORAGE_ENDPOINT: "{{ lookup('env', 'DJANGO_MINIO_STORAGE_ENDPOINT') }}"
+        DJANGO_MINIO_STORAGE_ACCESS_KEY: "{{ lookup('env', 'DJANGO_MINIO_STORAGE_ACCESS_KEY') }}"
+        DJANGO_MINIO_STORAGE_SECRET_KEY: "{{ lookup('env', 'DJANGO_MINIO_STORAGE_SECRET_KEY') }}"
```
brianhelba, line 21:
```suggestion
```

ansible/main.yml
```diff
@@ -0,0 +1,110 @@
...
+        DJANGO_STORAGE_BUCKET_NAME: "{{ lookup('env', 'DJANGO_STORAGE_BUCKET_NAME') }}"
+
```
brianhelba, line 23: Also need to set:
```
DJANGO_ALLOWED_HOSTS: <host variable>
DJANGO_DEFAULT_FROM_EMAIL: admin@<host variable>
DJANGO_STORAGE_BUCKET_NAME: <bucket variable>
```

brianhelba, line 23: @banesullivan We also need to know if there are credentials for Boto3 here, or whether it's done implicitly via an EC2 Instance Profile.

ansible/main.yml
```diff
@@ -0,0 +1,110 @@
...
+  - name: install pip + python, postgres headers 
+    apt: 
+      pkg:
+        - python3-pip
+        - python3-dev
```
brianhelba, line 28:
```suggestion
        - python3-pip
        - python3-dev
        - python3-setuptools
        - python3-venv
```
This will get the virtualenv packages for us.

brianhelba, line 28: @brianhelba Is `python3-dev` required?
https://github.com/girder/ansible-role-girder-celery/blob/master/tasks/main.yml#L14 doesn't use it.

ansible/main.yml
```diff
@@ -0,0 +1,110 @@
...
+      update_cache: yes
+      state: present
+    become: yes
+
+  - name: install virtualenv
+    pip:
+      name: 
+        - virtualenv
+    become: yes
+
```
brianhelba, line 39:
```suggestion
```

ansible/main.yml
```diff
@@ -0,0 +1,110 @@
...
+  - name: create virtualenv
+    shell:
+      cmd: /usr/local/bin/virtualenv {{ app_path }}/venv -p python3.8
```
brianhelba, line 42:
```suggestion
- name: Ensure virtualenv is using a recent pip version
  pip:
    name: "pip"
    # This is the first version with the new resolver
    version: ">=20.3"
    virtualenv: "{{ celery_virtualenv_path }}"
    # This implicitly creates a virtualenv, so do so using the Python 3 syntax
    virtualenv_command: "python3 -m venv"
  become: true
```

ansible/main.yml
```diff
@@ -0,0 +1,110 @@
...
+
+  - name: install pip requirements
+    shell: 
+      cmd: "{{ venv }}/bin/pip install -e .[dev] --find-links=https://girder.github.io/large_image_wheels"
+      chdir: "{{ app_path }}"
```
brianhelba, line 47:
```suggestion
  - name: install pip requirements
    pip:
      name: "{{ app_path }}"
      extra_args: "--find-links=https://girder.github.io/large_image_wheels"
      virtualenv: "{{ celery_virtualenv_path }}"
    become: true
```

ansible/main.yml
```diff
@@ -0,0 +1,110 @@
...
+
+  - name: run django migrations
+    django_manage:
+      command: migrate
+      app_path: "{{ app_path }}"
+      virtualenv: "{{ venv }}"
+    environment: "{{ watch_environment }}"
+
```
brianhelba, line 55: Also need to run `./manage.py collectstatic` here.

ansible/main.yml
```diff
@@ -0,0 +1,110 @@
+---
+- hosts: all
+  vars:
+    app_path: /home/ubuntu/RD-WATCH
```
brianhelba, line 4:
```suggestion
    app_path: "{{ ansible_facts['user_dir'] }}/RD-WATCH"
```
This avoids hardcoding the home dir.

ansible/main.yml
```diff
@@ -0,0 +1,110 @@
...
+    venv: "{{ app_path }}/venv"
```
brianhelba, line 5:
```suggestion
    venv: "{{ ansible_facts['user_dir'] }}/venv"
```

Let's not put the virtualenv inside the source repo.

ansible/main.yml
```diff
@@ -0,0 +1,110 @@
...
+    template:
+      src: "./gunicorn/gunicorn.service.j2"
+      dest: /etc/systemd/system/gunicorn.service
+    become: yes
+
+  - name: start gunicorn service
+    systemd:
+      name: gunicorn.service
+      enabled: yes
+      state: started
```
brianhelba, line 108:
```suggestion
      state: restart
```
To ensure code changes get picked up by the running webserver.

ansible/main.yml
brianhelba: On second thought, I think you should create a new root-level `requirements.txt` in this repo, matching what we used to have here: https://github.com/ResonantGeoData/ResonantGeoData/blob/0.1.2/requirements.txt

Then, you can reference that file in this step (you'll have to rewrite it, of course).

Also, if you use https://github.com/girder/ansible-role-girder-celery , that will automagically pick up the requirements file and use it.

watch/settings.py
```diff
@@ -37,6 +37,8 @@ def before_binding(configuration: Type[ComposedConfiguration]):
             # TODO: Log that we aren't using the expected GDAL wheel?
             pass
 
+        configuration.ALLOWED_HOSTS = ['watch-test.resonantgeodata.com']
```
brianhelba, line 40: You should be able to remove this, once it's set via environment variable.

ansible/main.yml
banesullivan: It should be done implicitly via an EC2 Instance Profile

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,38 @@
...
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      - name: Run playbook
+        env: # TODO: Hard code, "non-secret" values
+          INSTANCE_ID: ${{ secrets.INSTANCE_ID }}
+          DJANGO_DATABASE_URL: ${{ secrets.DJANGO_DATABASE_URL }}
+          DJANGO_SECRET_KEY: ${{ secrets.DJANGO_SECRET_KEY }}
+          DJANGO_EMAIL_URL: ${{ secrets.DJANGO_EMAIL_URL }}
+          DJANGO_STORAGE_BUCKET_NAME: ${{ secrets.DJANGO_STORAGE_BUCKET_NAME }}
+          DJANGO_CELERY_BROKER_URL: ${{ secrets.DJANGO_CELERY_BROKER_URL }}
+          DJANGO_ALLOWED_HOSTS: ${{ secrets.DJANGO_ALLOWED_HOSTS }}
```
venkatabhishek, line 33: We will need to set these secrets or hard-code them in playbook if not actually secret

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,38 @@
...
+          python-version: "3.8"
+      - name: Install Ansible
+        run: |
+          pip install --upgrade pip
+          pip install ansible
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
```
venkatabhishek, line 24: Convert to use ssm create session with private key

brianhelba (approved):

PR #10, opened 2021-08-12
use inventory group

venkatabhishek:


brianhelba:
@venkatabhishek What's the purpose of this change?

PR #14, opened 2021-08-18
remove docker development playbook

venkatabhishek:


brianhelba:
What about `ansible/.env.test`?

PR #19, opened 2021-08-19
New Region model for WATCH

### LINKED ISSUE #3: New Region model
Implement a Region model per https://infrastructure.smartgitlab.com/docs/pages/api_documentation.html#region-model

Then have an associated task on that model to populate Landsat/Sentinel Imagery from #2 

This model needs to store the start/end dates and the version tag

On save/update, reprocess overlapping satellite imagery in that time frame

banesullivan:
Resolve #3


banesullivan:
@matt-bernstein, is it necessary to have a `version` tag stored here? Looking at the example in https://infrastructure.smartgitlab.com/docs/pages/api_documentation.html#region-model, not all of the features have a version tag

matt-bernstein:
> @matt-bernstein, is it necessary to have a `version` tag stored here? Looking at the example in https://infrastructure.smartgitlab.com/docs/pages/api_documentation.html#region-model, not all of the features have a version tag

There is one semver tag per geojson file, not one per feature.

For region models (https://smartgitlab.com/TE/annotations/-/tree/master/D1/region_models), it's the `properties:version` field of the `Feature` with `properties:type == "region"`. It's a little confusing that a region and the sites it contains are stored together in a flat `FeatureCollection`, and you have to parse that field to tell the difference.

For site models (https://smartgitlab.com/TE/annotations/-/tree/master/D1/site_models), it's the `version` tag in the root of the file, which makes more sense.

banesullivan:
Thanks, @matt-bernstein!

>It's a little confusing that a region and the sites it contains are stored together in a flat FeatureCollection, and you have to parse that field to tell the difference.

That's incredibly confusing. I will update this to reflect that though

> For site models ...

I'm hoping I don't need to handle Site models any time soon

PR #42, opened 2021-12-10
Treat post_stac as "get or create"

banesullivan:
Related: 
- https://github.com/ResonantGeoData/ResonantGeoData/pull/633
- https://github.com/ResonantGeoData/ResonantGeoData/issues/629


AlmightyYakob:
Relies on client work to be implemented in https://github.com/ResonantGeoData/ResonantGeoData/pull/633, specifically in `create_file_from_url`. Once that's done, we can update the client version, and merge this.

banesullivan:
@AlmightyYakob, I think we need a fitler on STACFile to do the same check (see if a STACFile exists for the given ChecsumFile ID).

AlmightyYakob:
> @AlmightyYakob, I think we need a fitler on STACFile to do the same check (see if a STACFile exists for the given ChecsumFile ID).

You're right, I'll add that.

AlmightyYakob:
@banesullivan I've added that, so once that upstream PR is merged and incorporated here, that failing test should pass. Also, even though I've used the `swagger_auto_schema` decorator, the swagger page doesn't show the `file` field in the endpoint for stac file listing. I'm not sure why this is, but figured it shouldn't block this PR.

PR #44, opened 2021-12-10
Dry run of STAC Catalog iteration

banesullivan:
A dry run test of iterating over the STAC Catalogs. Both Sentinel and Landsat catalogs error after 10000 items, but the Landsat one gives an error message:

```
search_phase_execution_exception: [illegal_argument_exception] Reason: Result window is too large, from + size must be less than or equal to: [10000] but was [10010]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting.
```

The S3 static catalogs work fine.

----

This seems relevant: https://github.com/sat-utils/sat-api/issues/225


banesullivan:
@mcovalt, will you be able to look into this after finishing the deployment scrips, etc?

banesullivan:
This is working for everything but landsat/sentinel. Merging...

We can enable the LS/S2 in a follow up PR

PR #48, opened 2021-12-14
Add create_watch_client method

### LINKED ISSUE #47: Set correct production instance as api_url
@AlmightyYakob, could we set this up so that if using the `rgd_watch_client` plugin, it overrides the `api_url` by default?

AlmightyYakob:
Closes #47

This new method should be used in place of `create_rgd_client`.


AlmightyYakob:
@banesullivan lmk if documentation for using this new method should be added somewhere.

banesullivan:
I'll worry about documenting this when producing the demo materials. Thanks, @AlmightyYakob

PR #50, opened 2021-12-14
Debug LS and S2 STAC iteration

banesullivan:
Follow up to #44


banesullivan:
Update: we need to reduce our queries because stac-server has some serious limitations.

The problem and the error that we are experiencing (which index.max_result_window controls) is that stac-server is improperly using the elastic search scroll API. This seems like a serious flaw in stac-server's implementation and a significant limitation of using stac-server.

Specifically, from the Elastic Search API docs, there is this note:

> We no longer recommend using the scroll API for deep pagination. If you need to preserve the index state while paging through more than 10,000 hits, use the search_after parameter with a point in time (PIT).

ref https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html#scroll-search-results

This error is indeed missing functionality / a bug in stac-server itself and has already been filed with stac-server here: https://github.com/stac-utils/stac-server/issues/111

To work around this until stac-server can resolve the issue is partition the data using a filter on the DateTime (e.g. per day)  with https://api.stacspec.org/v1.0.0-beta.4/item-search/#operation/getItemSearch

banesullivan:
I've updated the LS and S2 ingest scripts to use pystac-client - we'll see it's able to iterate over the whole collection in a reasonable amount of time... if not, will try to parallelize in a follow-up PR as we parallelize the POST requests to RGD

scripts/landsat.py
```diff
@@ -1,17 +1,20 @@
+from datetime import datetime
+
 from watch_helpers import post_stac_items_from_server
 
-host_url = 'https://landsatlook.usgs.gov/stac-server/collections/landsat-c2l1/items'
+min_date = datetime(2013, 1, 1)  # Arbitrarily chosen
```
banesullivan, line 5: @matt-bernstein, any thoughts on a minimum ingest date for Landsat? 

I'm waiting to see how long it takes to ingest, so assuming the last 8 years of data doesn't take more than an hour or so to ingest, there shouldn't be much of a limitation on our end

scripts/watch_helpers/__init__.py
```diff
@@ -22,40 +23,57 @@ def iter_matching_objects(
...
-        if link['rel'] == 'self':
-            return link['href']
-    raise ValueError('No self link found')
+def handle_posts(iter_func, collection, dry_run, *args, **kwargs):
+    client = get_client(dry_run)
+    with concurrent.futures.ThreadPoolExecutor(max_workers=32) as executor:
+        future_to_url = {
+            executor.submit(client.watch.post_stac_file, url, collection): url
+            for url in iter_func(*args, **kwargs)
+        }
```
banesullivan, line 64: This is a problem. We need each thread to individually and concurrently use the iterator

banesullivan:
stac-server's response times are proving to be slow. We're seeing ~5 seconds to retrieve a single day's items. If we want to iterate over the last 5 years (arbitrarily chosen but we'll def want more data than that), it would take over 2.5 hours (365 * 5 * 5 / 60 / 60 = 2.535) if done serially.

PR #53, opened 2021-12-15
Ingest threaded posts

### LINKED ISSUE #52: Data ingestion takes too long
The worldview ingest routine timed out after 6 hours (GitHub's limit): https://github.com/ResonantGeoData/RD-WATCH/runs/4516298970?check_suite_focus=true

The bottleneck is the RGD server -- how can we speed this up? Perhaps parallelizing the post requests and spawning a huge pool of threads?

@mcovalt, any thoughts?

banesullivan:
Resolve #52

### DIFF
--- a/scripts/watch_helpers/__init__.py
+++ b/scripts/watch_helpers/__init__.py
@@ -67,7 +67,7 @@ def handle_posts(iter_func, collection, dry_run, *args, **kwargs):
     client = get_client(dry_run)
     with concurrent.futures.ThreadPoolExecutor(max_workers=n) as executor:
         future_to_url = {
-            executor.submit(client.watch.post_stac_file, client.watch.post_stac_file): url
+            executor.submit(client.watch.post_stac_file, url, collection): url
             for url in iter_func(*args, **kwargs)
         }
         for future in concurrent.futures.as_completed(future_to_url):
--- a/scripts/watch_helpers/__init__.py
+++ b/scripts/watch_helpers/__init__.py
@@ -63,8 +63,9 @@ def get_client(dry_run: bool = False):
 
 
 def handle_posts(iter_func, collection, dry_run, *args, **kwargs):
+    n = min(32, (os.cpu_count() or 1) + 4)
     client = get_client(dry_run)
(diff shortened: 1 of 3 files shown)


scripts/watch_helpers/__init__.py
```diff
@@ -60,6 +62,23 @@ def get_client(dry_run: bool = False):
...
+            for url in iter_func(*args, **kwargs)
+        }
+        for future in concurrent.futures.as_completed(future_to_url):
+            url = future_to_url[future]
+            try:
+                _ = future.result()
+            except Exception as exc:
+                print('%r generated an exception: %s' % (url, exc))
+            else:
+                print(f'Succeeded: {url}')
```
banesullivan, line 79: @mcovalt, would you please take a look at this... I'm getting request timeouts with this. I suspect n_workers=32 isn't the right way to configure this?

mcovalt: Request timeouts make me think the issue is now on the backend. We should be able to handle that many requests. We can add a PostgreSQL connection pooler and look into gevent.

banesullivan:
Update: we need to reduce our queries because stac-server has some serious limitations.

The problem and the error that we are experiencing (which index.max_result_window controls) is that stac-server is improperly using the elastic search scroll API. This seems like a serious flaw in stac-server's implementation and a significant limitation of using stac-server.

Specifically, from the Elastic Search API docs, there is this note:

> We no longer recommend using the scroll API for deep pagination. If you need to preserve the index state while paging through more than 10,000 hits, use the search_after parameter with a point in time (PIT).

ref https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html#scroll-search-results

This error is indeed missing functionality / a bug in stac-server itself and has already been filed with stac-server here: https://github.com/stac-utils/stac-server/issues/111

To work around this until stac-server can resolve the issue is partition the data using a filter on the DateTime (e.g. per day)  with https://api.stacspec.org/v1.0.0-beta.4/item-search/#operation/getItemSearch

banesullivan:
These changes are now in #50

PR #80, opened 2022-08-12
Add manage.py

mvandenburgh:


mvandenburgh:
@mcovalt I find `manage.py` more convenient than `django-admin` for running management commands in a django project, so if we could include it here as an additional way to run those that would be helpful - AFAIK it doesn't cause any issues with our current configuration.

mcovalt (approved):
Sounds good! I think adding this makes it more of a "normal" Django project which I'm in favor of.

PR #83, opened 2022-08-22
Add Django admin

mvandenburgh:
Adds the settings needed to enable the Django admin (taken mostly from `django-composed-configuration`), plus adds some auto-generated Admin classes.

### DIFF
--- a/django/src/rdwatch/admin.py
+++ b/django/src/rdwatch/admin.py
@@ -1,0 +1,50 @@
+from django.contrib import admin
+
+from .models import (
+    PredictionConfiguration,
+    Saliency,
+    SaliencyTile,
+    SatelliteImage,
+    Site,
+    TrackingConfiguration,
+)
+
+# general admin settings
+admin.site.site_header = "WATCH Admin"
+admin.site.site_title = "WATCH Admin"
+
+
+@admin.register(PredictionConfiguration)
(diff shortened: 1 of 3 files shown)


mcovalt:
I was hoping to avoid using Django Admin. Any particular reason to include it? My thinking to not include it is because:
1. It allows editing the data and this is primarily a read-only application
2. It's a bit difficult to maintain with respect to geospatial data

mvandenburgh:
I guess I'm just used to reaching for the Django Admin whenever I need to quickly preview something in the DB. Also, even for projects where the system is "read-only", I find it the most convenient way to tweak things when playing with test data (in development only, of course).
I'm not familiar with what it takes to get it working with GeoDjango / general geospatial stuff though. If you don't think it would be beneficial in general, I'm fine with closing this and setting this up locally for myself in `local_settings.py`.

mcovalt:
OK, yeah certainly if you find utility for it we should have an easy path to enable it. Perhaps a new environment variable `RDWATCH_ENABLE_DJANGO_ADMIN`?

And in terms of maintenance cost, let's just say no testing is required for the admin page in order to make a PR mergeable.

mvandenburgh:
@mcovalt I ended up just adding a `local_settings.py`, which has been working fine for me the last few months, so I don't think merging this is necessary.

PR #84, opened 2022-08-22
Add `django-extensions`

mvandenburgh:
Doing this primarily so I can use `shell_plus` to interact with the DB instead of writing SQL.


mcovalt:
Rather than including these kinds of dependencies for everyone, I would rather these developer-specific dependencies be managed in `requirements.dev.txt` ([see the second bullet here line](https://github.com/ResonantGeoData/RD-WATCH/blob/phase-ii/DEVELOPMENT.md#customization)).

mvandenburgh:
Oops I missed that, I'll move it to that file :+1:

PR #87, opened 2022-08-22
Ingest script

mvandenburgh:


django/src/rdwatch/management/commands/ingest.py
```diff
@@ -0,0 +1,272 @@
...
+
+        sensor_name = properties.get("sensor_name")
+
+        if not sensor_name:
+            assert label == Site.GROUND_TRUTH
+            saliency = None
+        else:
+            assert sensor_name
+
+            sensor_name = "L8" if sensor_name == "Landsat 8" else "S2"
```
mcovalt, line 148: This can be taken from `CropParse.sensor` with the file name from each GeoJSON Feature's `properties.source`.

django/src/rdwatch/management/commands/ingest.py
```diff
@@ -0,0 +1,272 @@
...
+
+            satellite_timestamp = properties.get(
+                "observation_date", properties.get("start_date")
```
mcovalt, line 151: This can be taken from `CropParse.timestamp` with the file name from each GeoJSON Feature's `properties.source`. It does include date and time information.

django/src/rdwatch/management/commands/ingest.py
```diff
@@ -0,0 +1,272 @@
...
+                site_geojson_file.parents[3]
+                / "_assets"
+                / "pred_saliency"
+                / f"{crop_string}_salient.tif"
+            )
+
+            # TODO: not sure why some crop strings don't have
+            # a corresponding saliency TIF; there's quite a few in
+            # the sample data that don't. If one is encountered,
+            # just skip it and move on to the next raster.
```
mcovalt, line 179: Awesome, thanks for spotting this. Either:
1. Our view of the data model is incorrect (each `Site` must be derived from a `Saliency`)
2. We're missing data

I'm banking on number 2, but I'll consult the WATCH team to make sure. For now, logging this is and continuing on the error like you've done is perfect.

django/src/rdwatch/models/site.py
```diff
@@ -15,6 +19,8 @@ class Site(models.Model):
     )
     saliency = models.ForeignKey(
         to="Saliency",
+        null=True,
+        blank=True,
```
mcovalt, line 23: Is this related to the [note here](https://github.com/ResonantGeoData/RD-WATCH/blob/98b94a11303be53937a417e14e3e780fc2f79789/django/src/rdwatch/management/commands/ingest.py#L176-L179)? Until we know for sure what's going on there, we should assume we're missing data rather than loosening our constraints. Right now, I can't mentally reconcile how a `Site` not derived from a `Saliency` can exist and our data model should reflect that.

django/src/rdwatch/management/commands/ingest.py
```diff
@@ -0,0 +1,258 @@
...
+        # directory of the source TIF given in the geojson.
+        crop_string = Path(properties["source"]).parent.name
+
+        # Parse the crop string, extracting the metadata encoded in it
+        crop_parse = CropParse.from_string(crop_string)
+
+        sensor_name = crop_parse.sensor
+
+        assert sensor_name
+        sensor_name = "L8" if sensor_name == "Landsat 8" else "S2"
```
mcovalt, line 143: I think `crop_parse.sensor` will already be `"L8"` or `"S2"`, not `"Landsat 8"` or `"Sentinel 2"`.

mvandenburgh:
@mcovalt I removed the dependency on `django-click` for this script and just used the native Django mechanism for creating `manage.py` scripts. The syntax isn't as clean, but mypy was reporting errors with `django-click` and I figured it's not worth spending time debugging it if it's going to be somewhat temporary anyways. I also fixed the remaining linting errors, so this should be good to go unless you have further comments.

mcovalt (approved):
Looks good! I've put a _bunch_ of work on a branch from this, so let's go ahead and get this merged in. After playing with this ingest and the data, I've changed the data model quite a bit and thus the ingestion.

PR #95, opened 2022-09-15
Drop `SatelliteImage` model

### LINKED ISSUE #94: Drop `SatelliteImage` model
The [`SatelliteImage` model](https://github.com/ResonantGeoData/RD-WATCH/blob/314088103c819aecefbe41f02c556c21d852dc7f/django/src/rdwatch/models/satellite_image.py) is populated during ingest to create a catalog of information required to read a remote image. This information is captured by scraping various STAC API servers at ingest time. This issue seeks to fully rely on these STAC API servers rather than duplicating their information.

## Currently

[During ingest](https://github.com/ResonantGeoData/RD-WATCH/blob/314088103c819aecefbe41f02c556c21d852dc7f/django/src/rdwatch/management/commands/ingest.py#L278-L308), STAC API servers are scraped to fetch relevant satellite images. The two modules [`rdwatch.utils.stac_search`](https://github.com/ResonantGeoData/RD-WATCH/blob/314088103c819aecefbe41f02c556c21d852dc7f/django/src/rdwatch/utils/stac_search.py) and [`rdwatch.utils.satellite_bands`](https://github.com/ResonantGeoData/RD-WATCH/blob/314088103c819aecefbe41f02c556c21d852dc7f/django/src/rdwatch/utils/satellite_bands.py) hold the logic to make these STAC API queries.

To read a satellite image tile, a request is made to [`/api/satellite-image/tile/<int:z>/<int:x>/<int:y>.webp`](https://github.com/ResonantGeoData/RD-WATCH/blob/314088103c819aecefbe41f02c556c21d852dc7f/django/src/rdwatch/urls.py#L29-L32) with the follow URL query parameters:
- **constellation**: One of `L8` or `S2` indicating whether we're pulling images from Landsat 8 or Sentinel 2
- **level**: One of `1C` or `2A` to indicate the processing level of the source imagery
- **spectrum**: Indicating the [common name of the band](https://github.com/stac-extensions/eo#common-band-names) the tile should provide
- **timestamp**: a time *around* when the satellite image was captured. This will not be precise, but it is expected the source imagery was captured within about an hour window around the provided timestamp.

The [view](https://github.com/ResonantGeoData/RD-WATCH/blob/314088103c819aecefbe41f02c556c21d852dc7f/django/src/rdwatch/views/tile.py#L98-L136) then redirects to [`/api/satellite-image/<int:pk>/tile/<int:z>/<int:x>/<int:y>.webp`](https://github.com/ResonantGeoData/RD-WATCH/blob/314088103c819aecefbe41f02c556c21d852dc7f/django/src/rdwatch/urls.py#L33-L37). This is done so the approximate timestamp will not invalidate cached requests to the same tile.

## Expected behavior

The `SatelliteImage` model is dropped and STAC API scraping during ingest is removed.

The API interface to `/api/satellite-image/tile/<int:z>/<int:x>/<int:y>.webp` stays the same.

In the associated view, the relevant STAC API will be queried with the two modules `rdwatch.utils.stac_search` and `rdwatch.utils.satellite_bands` at request time. The view then redirects to some internal URL that uniquely identifies the source imagery, perhaps even to the same path with a precise timestamp.

mvandenburgh:
Closes #94


django/src/rdwatch/views/tile.py
```diff
@@ -111,45 +114,38 @@ def satelliteimage_raster_redirect(
...
+    timestamp = datetime.fromisoformat(str(request.GET["timestamp"]))
+
+    # Calculate the bounding box from the given x, y, z parameters
+    bounds = mercantile.bounds(x, y, z)
+    bbox = (bounds.west, bounds.south, bounds.east, bounds.north)
+
+    bands = get_bands(constellation, timestamp, bbox)
+
+    try:
+        band: Band = next(bands)
```
mcovalt, line 128: Rather than grabbing the first band available, we should get the band filtered by:

- the requested processing level: `level`
- the requested spectrum (common band name): `spectrum`

From that filtered list, we should then sort it by:
- the closest timestamp to the requested timestamp. Since we allow a time "buffer" in our STAC search, it's possible (but unlikely) there will be multiple images returned within that buffer.
- preferring URIs with the ".tif" extension. For Sentinel 2, there's the possibility of getting a JP2000 or a Cloud Optimized GeoTIFF encoding for the exact same image. Cloud Optimized GeoTIFF is both cheaper and faster to generate tiles for.

mcovalt (changes requested):
This is excellent work and I'm excited to start trying it out! I added a comment to make a change we'll need before we can merge this in.

mcovalt (approved):
Awesome, works perfectly.

PR #102, opened 2022-10-13
Support `visual` spectrum in satellite raster image endpoint

mvandenburgh:


mcovalt:
Lets punt on this for a little while until after a refactor of [`phase-ii` as it stands now](https://github.com/ResonantGeoData/RD-WATCH/tree/f80c361efae9a845f73d141ed73d63b2fd5de506). I found while doing the WorldView imagery that trying to create a generic interface like it is now for all of Sentinel 2, WorldView, and Landsat 8 is [the wrong abstraction](https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction). There's just too much variability in the data. This PR further supports that by showing the Landsat 8 RGB processing is unique as well.

PR #109, opened 2022-11-02
Paginate Site Evaluations endpoint

### LINKED ISSUE #89: Paginate evaluations
Currently all evaluations are pulled into the sidebar. We'll need to paginate this. It would be cool to [virtualize the large list](https://vuejs.org/guide/best-practices/performance.html#virtualize-large-lists) and have that work in tandem with the pagination. The evaluation buttons are all the same size, so getting a count of all evaluations to buffer the scroll area should be doable.

mvandenburgh:
Closes #89 

Unfortunately, DRF's default pagination mechanism doesn't completely work here. DRF's `paginate_queryset` method returns a `list` instead of a `QuerySet`, and thus is intended to be called after all other queries have been made. Unfortunately, the end result of the queries made in the site evaluations endpoint is not a `QuerySet` but a `dict`, due to the use of `.aggregate()`. So, there doesn't appear to be a way to use DRF's native pagination methods here; instead, I've dropped down a level to Django's native `Paginator`, which the DRF one uses under the hood. 

I went ahead and configured pagination for DRF anyway in `settings.py`, and also refer to the `PAGE_SIZE` setting in the manual pagination code I wrote so it will be consistent with any DRF-paginated endpoints we have in the future.

### DIFF
--- a/django/src/rdwatch/views/site_evaluation.py
+++ b/django/src/rdwatch/views/site_evaluation.py
@@ -44,13 +44,8 @@ def site_evaluations(request: HttpRequest):
         queryset=SiteEvaluation.objects.order_by("-timestamp"),
     ).qs
 
-    # Pagination
-    assert api_settings.PAGE_SIZE, "PAGE_SIZE must be set."
-    paginator = Paginator(queryset, api_settings.PAGE_SIZE)
-    page = paginator.page(request.GET.get("page", 1))
-    queryset = page.object_list  # get the paginated queryset
-
-    queryset = queryset.annotate(
+    # Overview
+    overview = queryset.annotate(
         timemin=Min("observations__timestamp"),
         timemax=Max("observations__timestamp"),
     ).aggregate(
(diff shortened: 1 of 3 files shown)


django/src/rdwatch/views/site_evaluation.py
```diff
@@ -37,47 +39,72 @@ def get_responses(self, *args, **kwargs):
...
-                        min=ExtractEpoch("timemin"),
-                        max=ExtractEpoch("timemax"),
+    queryset = SiteEvaluationsFilter(
+        request.GET,
+        queryset=SiteEvaluation.objects.order_by("-timestamp"),
+    ).qs
+
+    # Pagination
+    assert api_settings.PAGE_SIZE, "PAGE_SIZE must be set."
+    paginator = Paginator(queryset, api_settings.PAGE_SIZE)
```
mvandenburgh, line 49: We could also accept a `page_size` query param here, and just treat `api_setting.PAGE_SIZE` as a maximum page size/default if the parameter is invalid. @mcovalt should I add that?

mcovalt: Good idea. Maybe in the future, but no need to include it here.

mcovalt (approved):
This looks good to me! @mvandenburgh be sure to review my changes in 0875f7d6495642b91fc9db962d74a032d2e2e61a. Also note the subsequent suppression of MyPy warning in ead97b51e2a4ea3a9195d9efe65839e08ce6d39c.

The commit 0875f7d6495642b91fc9db962d74a032d2e2e61a changes the behavior slightly so that `count`, `timerange`, and `bbox` cover the whole object list, rather than the paginated subset. Unfortunately, my change brings us to two queries that I hope we can reduce to just one in the future with a [Subquery](https://docs.djangoproject.com/en/4.1/ref/models/expressions/#subquery-expressions).

mvandenburgh:
Changes look good to me, thanks!

PR #122, opened 2022-11-30
Set up framework for UI filters, add Performer filter

mvandenburgh:


vue/tailwind.config.cjs
```diff
@@ -6,5 +6,8 @@ module.exports = {
       mono: ["Iosevka", "monospace"],
     },
   },
-  plugins: [],
+  plugins: [require("daisyui")],
+  daisyui: {
+    themes: ['lofi'],
```
mvandenburgh, line 11: I didn't put too much thought into this theme, but it seemed to match the existing theme of the app the best.

vue/src/mapstyle/rdwatchtiles.ts
```diff
@@ -10,16 +10,26 @@ import type {
...
+      "any",
+      [">", ["get", "timemax"], timestamp],
+      ["==", ["get", "timemin"], ["get", "timemax"]],
+    ],
+  ];
+
+  // Add any filters set in the UI
+  Object.entries(filters).forEach(([key, value]) => {
+    filter.push(["==", ["get", key], value]);
+  });
```
mvandenburgh, line 29: This is the core of the filtering framework i attempted to set up. it dynamically generates maplibre filters based on the `filters` object in the store. Currently, this only allows filtering by equality, which will need to be refactored when we get to filtering by ranges, etc.

PR #123, opened 2022-12-05
Allow setting page size

mcovalt:
A limit of "0" results in essentially no limit. Rather than handling this as a completely separate case, I just default to a really big integer when providing a limit of "0". This results in a really small diff, but may be confusing later on. @mvandenburgh, let me know if you have a better way of doing this.

### DIFF
--- a/django/src/rdwatch/views/site_evaluation.py
+++ b/django/src/rdwatch/views/site_evaluation.py
@@ -1,3 +1,5 @@
+import sys
+
 import django_filters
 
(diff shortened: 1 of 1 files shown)


mcovalt:
I'll merge, but definitely feel free to refactor later.

mvandenburgh:
I think ideally, we'd want to validate that `page_size` is a positive integer inside the DRF Serializer for the query params. But since we're not really using DRF at this time, I agree this is clean enough for now.

PR #141, opened 2023-01-31
Environment Variable Tile Server

BryonLewis:
Switches the front-end to using an environment variable for the tile server instead of the hardcoded URLS.

Note this will need to be set during the build process so that the production build has the proper variable.
I made it default to the new URL if it can't find the env anyways.
ALSO:
```
VITE_TILE_SERVER_URL="https://basemap.kitware.watch"
```
Will need to be added to the .env file.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -6,8 +6,6 @@ services:
   # RD-WATCH
   rdwatch:
     image: ghcr.io/resonantgeodata/rd-watch/rdwatch:latest
-    build: 
-      context: '.'
     depends_on:
       - postgresql
       - redis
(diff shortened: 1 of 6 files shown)


docker-compose.yaml
```diff
@@ -6,6 +6,8 @@ services:
   # RD-WATCH
   rdwatch:
     image: ghcr.io/resonantgeodata/rd-watch/rdwatch:latest
+    build:
```
mvandenburgh, line 9: What is the purpose of this change? It looks like it will change the semantics of the `image` property here (https://stackoverflow.com/a/34316224)

mvandenburgh (approved):

docker-compose.yaml
BryonLewis: thanks, that was left over from when I was doing local builds.  I'll remove it.

PR #145, opened 2023-02-08
Fix Pattern Settings Persistence

BryonLewis:
Updates the pattern settings so they persist between opening and closing the settings location.

### DIFF
--- a/vue/src/components/SettingsPanel.vue
+++ b/vue/src/components/SettingsPanel.vue
@@ -4,90 +4,110 @@ import type { Ref } from "vue";
 import { state } from "../store";
 import { updatePattern } from "../interactions/fillPatterns";
 
-const showSiteOutline = computed({ 
-  get() {  return state.filters.showSiteOutline || false },
-  set(val: boolean) {state.filters = { ...state.filters, showSiteOutline: val} }
+const showSiteOutline = computed({
+  get() {
+    return state.filters.showSiteOutline || false;
+  },
+  set(val: boolean) {
+    state.filters = { ...state.filters, showSiteOutline: val };
+  },
 });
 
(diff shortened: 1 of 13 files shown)


aashish24:
@BryonLewis, thanks for pushing this. The changes look reasonable, but I would have @mvandenburgh give his thumbs up. I haven't tried to install it locally.

vue/src/components/SettingsPanel.vue
```diff
@@ -5,11 +5,11 @@ import type { Ref } from "vue";
...
-const showSiteOutline: Ref<boolean> = ref(false);
-const groundTruthPattern: Ref<boolean> = ref(false);
-const otherPattern: Ref<boolean> = ref(false);
-const patternThickness = ref(1);
-const patternOpacity = ref(255);
+const showSiteOutline: Ref<boolean> = ref(state.filters.showSiteOutline || false);
+const groundTruthPattern: Ref<boolean> = ref(state.filters.groundTruthPattern || false);
+const otherPattern: Ref<boolean> = ref(state.filters.otherPattern || false);
+const patternThickness = ref(state.patterns?.patternThickness || 1);
+const patternOpacity = ref(state.patterns?.patternOpacity || 255);
```
mvandenburgh, line 11: Instead of allocating new refs, and having watchers updating the "real" values in the `state` object, why not make these `computed` and have a computed setter for modifying the `state.*` values directly?

BryonLewis: Good point, I was just patching the issue but this makes more sense to do.

mvandenburgh (approved):

aashish24:
@BryonLewis looks like some checks are failing.

PR #147, opened 2023-02-10
site-summary ingest endpoint

mvandenburgh:
Adds a new endpoint for ingesting `site-summary` JSONs.


mvandenburgh:
I'd like to refactor these endpoints to use pydantic/django-ninja, instead of the current approach which uses `TypedDicts` with several helper functions for extracting data from them. I'll make a follow-up PR for that.

BryonLewis (approved):
Pulled and tested.

More a comment/question than a change request:
Should these be tagged with some client side information that makes them different than a standard model run?  I.E, we have `parameters:{ground_truth: true}` on model runs that are ground truth.  Is there any reason these will need to be tagged separately to differentiate them from other runs?

PR #149, opened 2023-02-15
UI updates

BryonLewis:
Truncation of Performer/Region names if too big.
Truncation of Model Run Titles if too big
Displaying is locked to region+configuration Ids to prevent multiple areas being in view
Updated the loading mechanism so it doesn't reset the view
Added a loading progress bar when loading more model runs.
Added the total model run count to the UI.  NOTE: Total model run count is incorrect on back-end right now.


BryonLewis:
- Fix the image being null.
- Fix issues with deselecting items in the list.
- Add Region number and performer number to dropdowns
- Add background color and align the # of runs with the performer area.

BryonLewis:
Pull and test some toggling of model runs.  Hopefully it behaves now in a more predictable manner.

mvandenburgh (commented):
Looks great! One minor issue i noticed is that the tooltips for the performer/region dropdowns still appear with no text when you don't have anything selected 
tooltip

PR #150, opened 2023-02-16
Fix `count` field in model runs list endpoint

mvandenburgh:
Previously, we were returning the number of `regions` as the `count` field for the model runs endpoint. This PR fixes this to return the actual number of _model runs_ instead.


BryonLewis (approved):
Just pulled and tested with my new count and it works.

PR #151, opened 2023-02-24
Refactor ingest, render "site" polygons with `status` field as labels

mvandenburgh:
This PR does a couple things -

- Refactors ingest to use pydantic for parsing/validating jsons during ingest, instead of having `TypedDict`s and pulling things out manually. 
	- Eventually, I'd like to move away from the scattered approach we're currently taking with using DRF viewsets in some places, DRF `api_view`s in other places, as well as plain Django views in others, and move towards unifying our API to use [Django Ninja](https://django-ninja.rest-framework.com/) which natively integrates with Pydantic. 
- Retools the site-summary ingest to store site-summary geometries directly on the `SiteEvaluation.geom` column instead of creating a seperate `SiteObservation` for it.
- Add a `label` column to `SiteEvaluations` and populate it with the site/site-summary's  `status` field.
	- Update the front end to render these labels accordingly


mvandenburgh:
@BryonLewis this is ready for review. I would recommend reading through commit-by-commit, as i tried to separate things out cleanly per commit.

vue/src/mapstyle/rdwatchtiles.ts
```diff
@@ -96,14 +103,38 @@ export const buildSiteFilter = (
...
+  ["==", ["get", "label"], 11],
+  "#00ffff",
+  ["==", ["get", "label"], 12],
+  "#ff00ff",
+  ["==", ["get", "label"], 13],
+  "#00ff00",
+  ["==", ["get", "label"], 14],
+  "#ff69b4",
+  ["==", ["get", "label"], 15],
+  "#ffe4c4",
```
BryonLewis, line 137: Not needed for this PR but in the future we should probably build a typescript Type definition for these annotations/colors/labels.  I didn't know this before but probably basing it off of the lookups.json would work. Then we have a function which could generate this and the below text-labels.  It would be nice to have a direct tying of the color to the text displayed.

BryonLewis (approved):

vue/src/mapstyle/rdwatchtiles.ts
mvandenburgh: Agreed!

PR #158, opened 2023-03-14
Satellite image Display

BryonLewis:
Utilizes the S3 Sentinel images for visualizing timestamps:

- When selecting a Region it will request the timestamps for that given region's bounding box.  This may take a few seconds
- If the Region is larger than 1 degree in either the X/Y size it will display a warning.  I don't want the system to sit there trying to load timestamps for an abnormally large region because the backend would continue to try to grab data even if the user leaves the page or goes to another region.
- There is an indication of the satellite list loading in the same location as the regular loading bar.
     - image

- After the images have loaded a chip will be displayed showing the number of satellite images for the region.  Also the image icon will be enabled
     - image

- At this point the image icon will be enabled to toggle images on (if no region is selected or no images are found for a region it will remain disabled.
     - image
- The images displayed are synchronized with the time slider.  The closes timestamp is displayed in the chip as the user moves the timeslider.
- If the user closes the region the images will automatically disable until a new region is selected.
- within the settings there is an option to change the image opacity as well as an additional image checkbox which corresponds to the image icon.
    - image
- If the region is larger than 1 degree in X/Y direction a warning will come up informing the user that it is too large to get the satellite images.  The warning can be dimissed.
    - image

Issues for the future:

- First time it takes 30-90 seconds or more to load the timestamps.  On large regions, it may take even longer to load the timestamps.  That is why on regions > 1 degree X/Y it will display the warning message.
- If you jump to another region while the current one is in the process of loading it will continue to download data on the server in an attempt to get the timestamps.  This is something that should be resolved in the future.
- I've noticed a few times that the Access key for S3 has come up with warnings while testing.


mvandenburgh (approved):

BryonLewis:
Per the feedback from Aashish:

- There is now a small loading bar when loading satellite tiles:
    - image
- Additionally the satellite icon flashes while loading the tiles.

PR #159, opened 2023-03-14
Visualize region polygons

mvandenburgh:


django/src/rdwatch/views/model_run.py
```diff
@@ -226,7 +226,12 @@ def list(self, request: Request):
         # the database.
         if 'region' in request.query_params:
             aggregate |= queryset.defer('json').aggregate(
-                bbox=BoundingBoxGeoJSON('evaluations__geom')
```
mvandenburgh, line 229: I forgot we had added this optimization - we're not calculating the bounding box across all observations anymore, instead we're just using the site polygon. I've updated this logic to use the region polygon when possible, and only fall back on the site polygon if the region polygon doesn't exist (the site polygon is guaranteed to exist, but the region polygon can be NULL).

vue/src/components/SettingsPanel.vue
```diff
@@ -215,6 +224,21 @@ watch(hiddenCanvas, () => {
         >
       </label>
     </div>
+    <div class="form-control">
+      <label class="label cursor-pointer">
+        <img
+          ref="polygonImg"
```
BryonLewis, line 230: `polygonImg` is a ref to two locations (once for the region polygon and once for the site outline).  It means that only one of them is set as an actual image.  The site Outline is just an empty image of a specific size.  Not a big deal if we want to indicate site-outline to be an empty image and take advantage of the default border for images.  I just suggest we remove the reference from it or specify a different reference like the previous one `siteOutlineImg` if we want to tie it to a specific color.  These are simpler than the patterns so they could just be 'div' of size X with computed style or class indicating the border color as well.
image

BryonLewis (changes requested):
Just a small thing regarding the settings.
Everything else looks good and I'll approve as soon as I can demo it.

vue/src/components/SettingsPanel.vue
mvandenburgh: I updated it to just use a `div` with a grey border for the icon instead of an img/ref approach https://github.com/ResonantGeoData/RD-WATCH/pull/159/commits/a7443e86375f4fd82b60990d7ebbd988b070f0d6.

PR #160, opened 2023-03-17
Improved ingest error messages

mvandenburgh:
Currently, the ingest API returns 500 errors if an invalid site or region model JSON is posted. To fix this, we could catch these errors, format them in a useful way, and send them back in the endpoint response.

There's also another option. We're currently using `pydantic` to parse and validate these JSONs. `django-ninja` uses pydantic models for request/response serialization, and will automatically handle validation errors and send back an appropriate response for us, without needing any manual exception handling. So, I used this as an opportunity to convert the two ingest endpoints from DRF to django-ninja. The nice thing is that the rest of the endpoints can continue to use DRF or Django views, and we can convert those piecemeal over to `django-ninja` when we have time.

With these changes, an error message like this will be returned when, for example, a required field is missing in the site/region model:
```
{
  "detail": [
    {
      "loc": [
        "body",
        "site_model",
        "features",
        1,
        "geometry"
      ],
      "msg": "field required",
      "type": "value_error.missing"
    }
  ]
}
```

(This error indicates that the 2nd element in the site model's `features` array is missing the `geometry` field)

### DIFF
--- a/django/poetry.lock
+++ b/django/poetry.lock
@@ -256,6 +256,23 @@ python-versions = ">=3.7"
 Django = ">=3.2"
 
 [[package]]
+name = "django-ninja"
+version = "0.21.0"
+description = "Django Ninja - Fast Django REST framework"
+category = "main"
+optional = false
+python-versions = ">=3.7"
+
+[package.dependencies]
+Django = ">=2.2"
+pydantic = ">=1.6,<2.0.0"
+
+[package.extras]
+dev = ["pre-commit"]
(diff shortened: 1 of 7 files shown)


BryonLewis (approved):
I looked through it and nothing jumped out to my untrained eye.  Also pulled and ran it.  ?Hit the endpoints using that newer python script you gave me and it seemed to succeed.   Modified that script to corrupt the format and proper GDALErrors were returned.

PR #161, opened 2023-03-17
Configure `pytest` and add test for ingest

mvandenburgh:
Configures `pytest` and `pytest-django`, and adds a single test for ingest.

There's definitely more aspects of the system that need to be tested, this is just a start to get everything in place we need to be able to write more tests easily in the future.


mvandenburgh:
@BryonLewis can you try running the test suite locally using the new instructions https://github.com/ResonantGeoData/RD-WATCH/pull/161/commits/4f6cfdc463c95ea2e0bd9484bfa280d07918a74a, and let me know if it works for you?

mvandenburgh:
> > @BryonLewis can you try running the test suite locally using the new instructions [4f6cfdc](https://github.com/ResonantGeoData/RD-WATCH/commit/4f6cfdc463c95ea2e0bd9484bfa280d07918a74a), and let me know if it works for you?
> 
> ```
>   test: OK (61.97=setup[58.65]+cmd[3.32] seconds)
>   check-migrations: OK (23.10=setup[21.79]+cmd[1.32] seconds)
>   congratulations :) (85.13 seconds)
> ```
> 
> Might want a note about running `docker-compose run --rm poetry add tox` before being able to run the tests.
> 
> I don't know if it's this PR or another one, but we should update docker-entrypoint.sh to remove the automatic migration that occurs on `poetry` and `npm` docker-compose services.

Opened #164 to fix this.

PR #164, opened 2023-03-20
Docker images: only run django migrate if `django-admin` is installed

mvandenburgh:


BryonLewis (commented):
👍

BryonLewis (approved):
meant to approve before

PR #165, opened 2023-03-23
Fix errors in site model ingest

mvandenburgh:
This PR:

- Fixes some incorrect usage of Pydantic validation that was resulting in some 500 errors in the ingest endpoints (https://github.com/ResonantGeoData/RD-WATCH/pull/165/commits/afaeb292b830356261b4c05be89def7c85f8a0fc)
- Improves the error messages returned when a user attempts to ingest an invalid site/region model json (https://github.com/ResonantGeoData/RD-WATCH/pull/165/commits/aed6702733ab29ddc854195189cd26f6695f7518)
- Makes the parsing of comma-separated lists during site model ingest cleaner by casting them to a list of strings at the start of ingestion (https://github.com/ResonantGeoData/RD-WATCH/pull/165/commits/08ae93cbb84defb8d78f05cbbe46f0ac3dac6e9c)
- Removes support for comma-separated list for region models (these are not supported by the spec) https://github.com/ResonantGeoData/RD-WATCH/pull/165/commits/eec82ca18df4dbabfe5d7a8a4e6df692e82dcfcf
- Adds a new test for site model ingest, and cleans up some minor issues exposed by it (https://github.com/ResonantGeoData/RD-WATCH/pull/165/commits/0a77b2ac84b0af0871424c5b61127988f2c966b4, https://github.com/ResonantGeoData/RD-WATCH/pull/165/commits/40248e9dd8b1d9b8c3880f4babdb0b4ae132cca5, https://github.com/ResonantGeoData/RD-WATCH/pull/165/commits/cbd67def8eafb3b0dc2d8341b08e01161dccf2de)

- Refactors the site model ingest to be in line with what the spec expects https://github.com/ResonantGeoData/RD-WATCH/pull/165/commits/f75de8139f087c0c529b2dac8d254abdd38cdbe4
	- This part is particularly important, since I discovered that we have been making some incorrect assumptions about site models in both the ingest routine and in our data model since the beginning. Until now, we were storing each `observation` in a site model as a separate `SiteObservation` in the database, each with its own `MultiPolygon` geometry. However, this is incorrect - a single `observation` object in a site model actually maps to several `SiteObservations`. The `current_phase` field can contain multiple phases (in the form of a comma-separated list), each of which represents a different observation; and the nth phase in this list maps to the nth `Polygon` within the `MultiPolygon` for this `observation`. 
- To fix this, I
	- Changed the `SiteObservation` `geom` column to be a `PolygonField` instead of a `MultiPolygonField`.
	- Changed the site model ingest code to be in line with the facts above.
	- Wrote a custom database migration that will convert all the existing `SiteObservation` `geom` columns to a `Polygon`. This comes with some slight complications:
    	- `SiteObservations` whose `MultiPolygon` contained only a single `Polygon` are straightforward to convert. Practically, the vast majority of the `SiteObservations` fit into this category.
    	- `SiteObservations` whose `MultiPolygons` contained multiple `Polygons` are more difficult to convert. We _could_ have an even more complicated migration that handles this case; but considering that there's only a handful of these in the system, I think it's easier to just delete them and manually re-ingest later.

### DIFF
--- a/django/src/rdwatch/migrations/0005_alter_siteobservation_geom.py
+++ b/django/src/rdwatch/migrations/0005_alter_siteobservation_geom.py
@@ -1,0 +1,97 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+import django.contrib.gis.db.models.fields
+from django.contrib.gis.geos import MultiPolygon
+from django.db import migrations
+
+if TYPE_CHECKING:
+    from django.contrib.gis.db.backends.postgis.schema import PostGISSchemaEditor
+    from django.db.migrations.state import StateApps
+
+
+def migrate_observation_geom(apps: StateApps, schema_editor: PostGISSchemaEditor):
+    SiteEvaluation = apps.get_model('rdwatch', 'SiteEvaluation')  # noqa: N806
+    SiteObservation = apps.get_model('rdwatch', 'SiteObservation')  # noqa: N806
+
+    # Set of SiteEvaluation ids that we want to delete
+    evals_to_delete: set[int] = set()
+
+    for observation in SiteObservation.objects.iterator():
+        if len(observation.geom_old) > 1:
+            # If this MultiPolygon contains more than one Polygon, this
+            # site needs to be reingested
+            evals_to_delete.add(observation.siteeval.id)
+            observation.delete()
+        else:
+            # Otherwise, extract the one Polygon from the MultiPolygon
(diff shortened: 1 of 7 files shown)


BryonLewis (approved):
pulled, did the migration, and ran tests.  Did a cursory look over the changes and they look fine (although I don't have a detailed understanding of the ingest formats).

PR #167, opened 2023-03-24
Satellite Keyboard Controls, Annotation Styling, Pattern Density

BryonLewis:
- Modifies the Left/Right arrow keys so that they skip 1 day at a time.  If the satellite images are turn on it will instead jump to the next image in the list that is at least a day away.
- Satellite loading is throttled using lodash to hopefully prevent issues of trying to load too many timestamps at once
- The fade in/out for the satellite images has been reduced from 300ms to 0.  There is still a brief moment in time in which they disappear and a new one loads.  I'll look into removing that in the future.
- Pattern density controls have been added to the ground truth pattern for changing the density.  Since maplibre uses an automatic scaling system this allows you to choose from 5 densities which change the size of the underlying image to increase the perceived density of the lines.
- Refactored the AnnotationStyling into its own file where all of the observations and colors can be stored in one location and updated as needed instead of relying on flipping back and forth through multiple sections.


vue/vite.config.ts
```diff
@@ -23,10 +23,10 @@ export default defineConfig(({ mode }) => {
     plugins: [vue()],
     server: {
       host: "0.0.0.0",
-      port: 9000,
+      port: 8080,
       proxy: {
         "/api": {
-          target: "http://django:80",
+          target: "http://localhost:8000",
```
mvandenburgh, line 29: Accidental commit?

mvandenburgh (commented):
I noticed the new pattern densitry setting kind of breaks when you increase it too much. Also, the label needs to be updated (currently there's two "Pattern Thickness" labels"

Peek 2023-03-27 14-35

vue/vite.config.ts
BryonLewis: yup, I'll revert that.

PR #170, opened 2023-03-31
Refactor Satellite timeStamp requests

BryonLewis:
Refactors the request for satellite timestamps to do both sentinel and world-view at the same time.

- Instead of returning a flat list of timestamps it returns a list of objects (sorted by timestamp) which indicate each timestamp's source, collection, cloudcover).
- The front end has been modified to use this new data structure for the timestamps.
- There is a cloud cover filter and checkboxes to filter between S3 and WorldView Imagery.

Tips:

- US0005 is a decent sample region to look at to see the images.

Issues:

- Worldview Imagery takes a **LONG** time to download from the server if the region is decently large.  You can watch the console output and see the range requests happening.  I don't know enough about the COGReader and if there is a better way to do this.  I feel like it is trying to download/gather too much data based on the tile for some reason.
- So long in fact that there are 503 errors and some of the tiles just return 404 errors

### DIFF
--- a/django/src/rdwatch/utils/worldview_processed/stac_search.py
+++ b/django/src/rdwatch/utils/worldview_processed/stac_search.py
@@ -78,5 +78,4 @@ def worldview_search(
         headers={'x-api-key': environ['RDWATCH_SMART_STAC_KEY']},
     )
     with urlopen(request) as resp:
-        data = resp.read()
-        return json.loads(data)
+        return json.loads(resp.read())
--- a/django/src/rdwatch/utils/worldview_processed/stac_search.py
+++ b/django/src/rdwatch/utils/worldview_processed/stac_search.py
@@ -69,9 +69,7 @@ def worldview_search(
     else:
         time_str = f'{_fmt_time(timestamp)}Z'
     params['datetime'] = time_str
-    params['collections'] = [
-        "ta1-wv-acc-1",
-    ]
+    params['collections'] = ["ta1-wv-acc-2"]
     params['page'] = page
     params['limit'] = 100
     request = Request(
--- a/django/src/rdwatch/utils/worldview_processed/stac_search.py
+++ b/django/src/rdwatch/utils/worldview_processed/stac_search.py
@@ -69,7 +69,7 @@ def worldview_search(
     else:
         time_str = f'{_fmt_time(timestamp)}Z'
     params['datetime'] = time_str
-    params['collections'] = ["ta1-wv-acc-2"]
+    params['collections'] = ['ta1-wv-acc-2']
     params['page'] = page
     params['limit'] = 100
     request = Request(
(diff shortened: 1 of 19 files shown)


django/src/rdwatch/utils/worldview_processed/stac_search.py
```diff
@@ -78,4 +78,5 @@ def worldview_search(
         headers={'x-api-key': environ['RDWATCH_SMART_STAC_KEY']},
     )
     with urlopen(request) as resp:
-        return json.loads(resp.read())
+        data = resp.read()
+        return json.loads(data)
```
mvandenburgh, line 82: In the interest of minimizing redundant diffs, can this change be reverted? (unless it's intentional; I'm guessing it was broken out into a variable to assist in debugging though?)

mvandenburgh (commented):
One small UI bug I noticed when playing with this-

bug

If you deselect both S2 and WorldVIew in the UI, you are no longer able to disable the imagery by clicking the icon.

Also we'll want to have a well-defined schema for the response format instead of returning an opaque JSON object, not only to enhance readability but also so that the eventual Swagger API docs can be autogenerated. That's still a ways off though so I'm fine with merging this as-is.

django/src/rdwatch/utils/worldview_processed/stac_search.py
BryonLewis: Yeah that was some testing stuff that can be removed, thanks for catching that.

BryonLewis:
>Also we'll want to have a well-defined schema for the response format instead of returning an opaque JSON object, not only to enhance readability but also so that the eventual Swagger API docs can be autogenerated. That's still a ways off though so I'm fine with merging this as-is.

I'm hoping at some point this data is in a task populated table instead of requiring cascading requests to be cached.  That should give us some structure to it.

  When I make an update I'm probably going to turn off worldview by default (it's inconsistent in how it loads and it's behavior).  The way the worldview imagery is stored makes it fail when or take an insanely long time when asking for an entire region.  It will need further discussion but we may end up with something where we utilize the S2 imagery and this endpoint for full region satellite views, and then we utilize the smart-stac server for WorldView and S2 imagery if you want to view specific sites.

vue/src/store.ts
```diff
@@ -58,10 +68,22 @@ export const state = reactive<{
     satelliteBounds: [],
     imageOpacity: 0.75,
     loadingSatelliteImages: false,
+    satelliteTimeSource: 'S2',
+    cloudCover: 100,
+    satelliteSources: ['S2'],
```
mvandenburgh, line 73: Is worldview not included here because we're still having issues with loading tiles?

BryonLewis: Yeah, I don't want it to default to on, I only want people turning it on if they know what they're doing.

PR #172, opened 2023-04-11
Use SMART Program STAC server for S2/L8, make Accenture version configurable w/ env var

mvandenburgh:
- Switches the S2/L8 endpoint to use the SMART program's STAC API instead of the public S2/L8 APIs.
	- As part of this, since we're now using the same endpoint for both S2 and L8, I refactored the two search functions into one.
- Make accenture imagery version configurable via an environment variable


django/src/rdwatch/utils/stac_search.py
```diff
@@ -53,35 +56,23 @@ def _fmt_time(time: datetime):
...
 
-def sentinel_search(
+def stac_search(
+    source: Literal['S2', 'L8'],
     timestamp: datetime,
     bbox: tuple[float, float, float, float],
     timebuffer: timedelta | None = None,
     page: int = 1,
 ) -> Results:
-    url = 'https://earth-search.aws.element84.com/v0/search'
```
BryonLewis, line 84: Can we leave this URL as a comment, I remember Matt saying after the program we may want to revert back to public servers for serving images.

BryonLewis (approved):
Pulled and tested.  I'll update in the future to enable some L8 imagery.

Can you add the new env variable to the template.env, just so in the future we know what it should be or if anyone else pulls the code?

PR #174, opened 2023-04-11
Observation Images

BryonLewis:
Plan:

Server Side:

- Add S3FileField to Site Observations data with a default value of Null
- Create a task endpoint which includes a parameter for the site ObservationId
    - This task will load up all observations for the site and determine the image type (S3/WorldView) and download the image for the 'collective' (the average bounding box of all images).
    - It then stores that image in the S3FileField for the site observation so it can be passed back to the client side
- The updated site observations should include a reference to the S3FileField URL that is used.

Client Side:

- Ability to select multiple individual site observations
- When a site observation is selected we can kick of a task for each site observation to download the satellite images for it.
- These images can then all be turned on and filtered once they are loaded.
- System for caching images based on current time and having a list of enabled images
- System for displaying the images on the map over the bounding box that they represent.


BryonLewis:
More details about the client side implementation:

- Probably going to need a secondary panel to show selected observations.
- Other panel should indicate the timerange of the observation, number of no_activity, active_construction, no_construction fields.
- Indicate the number of images downloaded and ready
- Provide a button to hit the post endpoint to get the images
- Toggle to turn the images on for the area if they are all complete.

BryonLewis:
Next up:

- Make sure the returned minio URLs reference the base siteURL (localhost) instead of the network container host (minio)
- finalize the image retrieval and setup
- Switch from using 'video' as the field to 'sattelite_image' or something more descriptive
- Prevent it from attempting to download images that already exist
- Once images are stored and referenced work on the client side to display them.

BryonLewis:
A few things I didn't get done with the previous comment:

- Removal of selected sites from the list
- Auto deselect sites when changing region or toggling model runs
- Toggle to deselect site observations
- Skip to next/previous image timestamp buttons for site observations that have their images on


Future Update:
- enable downloading S3 and L8 Images
- Create filtering to be able to filter the S3 and L8 images
- Update 'video' reference in DB to be 'image' instead
- Possibly add cloudcover information to each site images if they exist

BryonLewis:
Next up:

- Adding a BBox scale to get more area around a site
- Adding playback control to initiate playback of the system
- Possible fade in/out of the system when changing between times.

PR #180, opened 2023-04-20
Reconfigure RDWatch CLI to get BBoxes directly

BryonLewis:
This PR will reconfigure the RDWatch CLI to get bboxes directly instead of trying to convert to tiles and stitching the area together.

Using the argument `--tile` will use the old behavior of returning a very large image using the tile method of getting pixels.  Without that argument, it will default to using the bounding box instead.

### DIFF
--- a/cli/src/rdwatch_cli/__init__.py
+++ b/cli/src/rdwatch_cli/__init__.py
@@ -119,7 +119,7 @@ async def image(
             img = await create_image(client, bbox, time, worldview=worldview)
             img.save(output)
         except ImageNotFound:
-            raise click.ClickException(
+            print(
                 'No image found for the given bounding box and time'
             )
         except ServerError:
@@ -191,7 +191,7 @@ async def movie(
             ]
             imgs[0].save(output, save_all=True, append_images=imgs)
         except ImageNotFound:
-            raise click.ClickException(
+            print(
                 'No image found for the given bounding box and time'
             )
         except ServerError:
--- a/cli/src/rdwatch_cli/__init__.py
+++ b/cli/src/rdwatch_cli/__init__.py
@@ -119,9 +119,7 @@ async def image(
             img = await create_image(client, bbox, time, worldview=worldview)
             img.save(output)
         except ImageNotFound:
-            print(
-                'No image found for the given bounding box and time'
-            )
+            print('No image found for the given bounding box and time')
         except ServerError:
             raise click.ClickException('Server down (try again later)')
 
@@ -191,8 +189,6 @@ async def movie(
             ]
             imgs[0].save(output, save_all=True, append_images=imgs)
         except ImageNotFound:
-            print(
-                'No image found for the given bounding box and time'
-            )
+            print('No image found for the given bounding box and time')
         except ServerError:
             raise click.ClickException('Server down (try again later)')
--- a/cli/src/rdwatch_cli/__init__.py
+++ b/cli/src/rdwatch_cli/__init__.py
@@ -5,7 +5,6 @@ from pathlib import Path
 
 import aiohttp
 import click
-
 from rdwatch_cli import api
 from rdwatch_cli.config import get_credentials, set_credentials
 from rdwatch_cli.exceptions import ImageNotFound, ServerError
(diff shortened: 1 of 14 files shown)


cli/src/rdwatch_cli/api/image.py
```diff
@@ -16,6 +16,39 @@ class WebpTile:
     image: Image.Image
 
 
+@dataclass
+class ObsImage:
```
mvandenburgh, line 20: imo `ObsImage` is too unclear for someone not familiar with the naming terminology for our internal data models.
```suggestion
class ObservationImage:
```

django/src/rdwatch/views/tile.py
```diff
@@ -184,6 +185,79 @@ def vector_tile(
...
+        )
+
+    query_params = {
+        **request.GET.dict(),
+        'timestamp': datetime.isoformat(precise_timestamp),
+    }
+
+    return HttpResponsePermanentRedirect(
+        reverse('satellite-bbox') + f'?{urlencode(query_params)}'
+    )
```
mvandenburgh, line 258: I think at the minimum it would be good to factor out the common code between this endpoint and the original `z/x/y/` endpoint into a common function, if possible. I'm less sure about this, but maybe a more ideal solution would be to remove the `z/x/y/` endpoint, and have a single endpoint that takes as query parameters {`z`, `x`, `y`} OR {`bbox` }. And for now we can make the `z/x/y` redirect to the new one until the frontend is updated accordingly. But I'm conflicted on whether that is better, since the `z/x/y/` url pattern is common in GIS systems I believe. Any thoughts on this? In general, I fear that we're adding too many similar endpoints for satellite images and the API is getting bloated.

cli/src/rdwatch_cli/__init__.py
```diff
@@ -172,8 +189,6 @@ async def movie(
             ]
             imgs[0].save(output, save_all=True, append_images=imgs)
         except ImageNotFound:
-            raise click.ClickException(
-                'No image found for the given bounding box and time'
-            )
+            print('No image found for the given bounding box and time')
```
mvandenburgh, line 192: Why do we no longer need an exception here?

mvandenburgh (commented):
This works correctly for me, after running it on the commands you shared on slack :+1:. Just left a couple comments on the code itself.

django/src/rdwatch/views/tile.py
BryonLewis: I don't know if we can convince maplibre to use query parameters instead of paths when generating URLS.  If we do it may be a non-standard thing that people may find awkward as you mentioned.  I'd have to look into that to see if it is possible.  I do agree with refactoring the commonality between the endpoints if possible.

cli/src/rdwatch_cli/__init__.py
BryonLewis: Without it, the tile version will crash when creating a movie if a single image for a specific x/y/z location isn't found.  I'd much rather have it log and skip the missing the image (maybe I add in the timestamp/bbox) then not create a movie.

BryonLewis: Also, I'm not a fan of the current vocab of the endpoints.  'visual' means worldview is quite a leap, but I didn't want to alter the existing stuff so I used the same pattern.

django/src/rdwatch/views/tile.py
BryonLewis: Also, I'm not a fan of the current vocab of the endpoints.  'visual' means worldview is quite a leap, but I didn't want to alter the existing stuff so I used the same pattern.

BryonLewis:
I refactored the requests for bbox/tiles for S2/L8 and worldview to utilize the same function.  Let me know if you think there should be an alternative way to do this.

django/src/rdwatch/views/tile.py
mvandenburgh: Agreed on the endpoint names, I think that's worth revisting once all of the demo code is reviewed and merged.

cli/src/rdwatch_cli/__init__.py
mvandenburgh: Ah I see, yeah I agree that it's better to skip the missing image instead of crashing. In that case, can we use `click.echo(...)` here instead of print?

mvandenburgh (approved):
LGTM once the dict key issue is fixed :+1:

mvandenburgh (approved):

PR #181, opened 2023-04-21
Basic MinIO configuration and setting up new table

BryonLewis:
Basic Minio configuration which uses docker-compose.override.yml to launch minio locally and set the environment variables.  This will change a bit when a proper Django ComposeConfiguration is implemented to use it.

The second part is getting the basic table set up for the siteImages.

SiteImages will be a list of images containing the image, the SiteEval foreign key, and siteObs foreign key if the time matches up with a known site observation time.  If not it will remain null.  I also added fields to connect the cloudcover for the time and the Image source ('S2', 'L8', 'WV').

Now when a task is set off it will gather all of the images associated with the observations first and then it will grab all of the images between the start/end times that may not be associated with an observation.


django/src/rdwatch/models/site_image.py
```diff
@@ -0,0 +1,39 @@
...
+        help_text='The site evaluation associated with this observation.',
+        db_index=True,
+    )
+    siteobs = models.ForeignKey(
+        to='SiteObservation',
+        on_delete=models.CASCADE,
+        help_text='The site Observation associated with this Image if it exists.',
+        null=True,
+        db_index=True,
+    )
```
mvandenburgh, line 19: I think we can just have a single foreign key to `SiteObservation`, and if we need to access the `SiteEvaluation` we can do a join.

BryonLewis: Correct me if I'm wrong, but for some images a `siteObservation` may not exist?  It's why I made it nullable.  The movement of this to a separate table was to get images that may be outside of the list of siteObservations.

mvandenburgh: Ah, right, I forgot there are some instances like that. This makes sense then

PR #183, opened 2023-04-24
Observation image task

BryonLewis:
- Configures MinIO storage using docker-compose.override.yml
- Table setup for SiteImages.  Includes Image, Timestamp, cloudcover, siteEval, siteObs (If Exists) and source ('WV', 'S2', 'L8)
    - Should auto delete the image in S3/MinIO if the table row is deleted
- Modified the default siteObservation to return the individual timestamp as well as a max/min time range.
- Migration for the new table
- Created `get_site_observation_images` tasks
    - If a siteImage exists for the timestamp, source and siteEval Id it will delete the existing image and replace it with the new one.
    - There is a second pass where it adds in images that aren't associated with a specific site observation
    - These are labelled slightly difference and don't include a foreign key for the siteobs_id
- Updated the `evaluations/<int:pk>` endpoint so that it returns an additional property called 'images' which is a list of the images that are downloaded and associated with the site observation.  There are some serializer updates to help with the creation of this list.


django/src/rdwatch/models/site_observation.py
```diff
@@ -14,13 +14,11 @@ class SiteObservation(models.Model):
     siteeval = models.ForeignKey(
         to='SiteEvaluation',
         on_delete=models.CASCADE,
-        help_text='The site evaluation associated with this observation.',
         db_index=True,
     )
     label = models.ForeignKey(
         to='ObservationLabel',
         on_delete=models.PROTECT,
-        help_text='Site observation classification label',
```
mvandenburgh, line 23: I don't think they're particularly helpful descriptions, but any particular reason to remove these?

django/src/rdwatch/tasks/__init__.py
```diff
@@ -1,6 +1,275 @@
...
+        for band in captures:
+            if band.level.slug == '2A' and band.spectrum.slug == 'visual':
+                tempCaptures.append(band)
+        captures = tempCaptures
+    # logger.warning(f'Captures: {len(captures)}')
+    if not captures:
+        return None
+    closest_capture = min(captures, key=lambda band: abs(band.timestamp - timestamp))
+
+    return closest_capture
```
mvandenburgh, line 95: Couldn't this function just call `get_range_captures` instead of repeating the same logic?

django/src/rdwatch/tasks/__init__.py
```diff
@@ -1,6 +1,275 @@
...
+
+
+def get_closest_capture(
+    bbox: tuple[float, float, float, float],
+    timestamp: datetime,
+    constellation: str,
+    worldView=False,
+):
+    timebuffer = timedelta(days=1)
+    captures = None
```
mvandenburgh, line 78: This initial assignment isn't necessary, as `if`/`elif`/`else` do not begin a new name scope. Additionally, this will make the type of the variable `list[WorldViewProcessedCapture] | None` instead of `list[WorldViewProcessedCapture]`, which would prevent a type-checker from catching certain type errors if we were using one.

django/src/rdwatch/tasks/__init__.py
```diff
@@ -1,6 +1,275 @@
...
+
+
+def get_range_captures(
+    bbox: tuple[float, float, float, float],
+    timestamp: datetime,
+    constellation: str,
+    timebuffer: timedelta,
+    worldView=False,
+):
+    captures = None
```
mvandenburgh, line 105: Same as above comment about scoping.

django/src/rdwatch/tasks/__init__.py
```diff
@@ -1,6 +1,275 @@
...
+    bbox: tuple[float, float, float, float],
+    timestamp: datetime,
+    constellation: str,
+    worldView=False,
+):
+    capture = get_closest_capture(bbox, timestamp, constellation, worldView)
+    # logger.warning(f'Closest Capture is: {capture}')
+    if capture is None:
+        return None
+    bytes = None
```
mvandenburgh, line 134: Same as above comment about scoping.

django/src/rdwatch/tasks/__init__.py
```diff
@@ -1,6 +1,275 @@
...
+    site_observations = SiteObservation.objects.filter(siteeval=site_eval_id)
+    transformer = Transformer.from_crs('EPSG:3857', 'EPSG:4326')
+    found_timestamps = {}
+    min_time = datetime.max
+    max_time = datetime.min
+    max_bbox = [float('inf'), float('inf'), float('-inf'), float('-inf')]
+    matchConstellation = ''
+    baseSiteEval = None
+    # First we gather all images that match observations
+    for observation in site_observations:
```
mvandenburgh, line 159: I think we should use `.iterator()` on this queryset 
https://docs.djangoproject.com/en/4.2/ref/models/querysets/#iterator
```python
    for observation in site_observations.iterator():
```

There's no max limit on the number of observations in a site, and I believe i've seen some with thousands; a site with thousands or tens of thousands of observations, each with its own complex geojson (multi)polygon, could blow up the memory on our worker.

django/src/rdwatch/tasks/__init__.py
```diff
@@ -1,6 +1,275 @@
...
+                else:
+                    SiteImage.objects.create(
+                        siteeval=observation.siteeval,
+                        siteobs=observation,
+                        timestamp=observation.timestamp,
+                        image=image,
+                        cloudcover=cloudcover,
+                        source=baseConstellation,
+                    )
+            os.remove(output)
```
mvandenburgh, line 218: Can we use a [`TemporaryFile`](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryFile) here instead of managing the file's lifecycle ourselves?

django/src/rdwatch/tasks/__init__.py
```diff
@@ -1,6 +1,275 @@
...
+                    existing.save()
+                else:
+                    SiteImage.objects.create(
+                        siteeval=baseSiteEval,
+                        timestamp=capture.timestamp,
+                        image=image,
+                        cloudcover=cloudcover,
+                        source=baseConstellation,
+                    )
+            os.remove(output)
```
mvandenburgh, line 275: Same question here about tempfile.

django/src/rdwatch/urls.py
```diff
@@ -50,6 +50,10 @@
         views.satelliteimage_visual_tile,
         name='satellite-visual-tiles',
     ),
+    path(
+        'observations/<int:pk>/get-images',
```
mvandenburgh, line 54: What do you think about using `generate-images` here instead of `get-images`? When I see `get-images`, I imagine that'd be a read-only GET endpoint.

django/src/rdwatch/models/site_observation.py
BryonLewis: CI for linting was fighting with pre-commit and I think flake8 or black were arguing with line length and causing new migrations so I just deleted the text.

django/src/rdwatch/tasks/__init__.py
BryonLewis: Yeah there could be some refactoring to combine some of the logic.  Range capture returns a list of captures whereas closest capture returns a single close capture or None.  The common logic could be refactored into a single function or the result of closes can attempt to find the closest value.

django/src/rdwatch/tasks/__init__.py
BryonLewis: good call

django/src/rdwatch/urls.py
BryonLewis: That's a better name.

PR #187, opened 2023-04-27
Add RabbitMQ to docker-compose + new env var for Celery Broker

mvandenburgh:


mvandenburgh:
@BryonLewis these are changes needed to accommodate the AWS deployment. Will this interfere with your PRs?

BryonLewis:
> @BryonLewis these are changes needed to accommodate the AWS deployment. Will this interfere with your PRs?

A cursory look doesn't look like it will impact anything.  Do we want to add a rabbitMQ service to the docker-compose to make deployment as similar to local dev as possible?
Also adding the new environment variable to the template.env and docker-compose to pass it into django?

mvandenburgh:
I reverted the changes to the Dockerfile, as I realized we can avoid having separate images for web app + worker. This is advantageous because it will ensure they stay synced up.
 
> > @BryonLewis these are changes needed to accommodate the AWS deployment. Will this interfere with your PRs?
> 
> A cursory look doesn't look like it will impact anything. Do we want to add a rabbitMQ service to the docker-compose to make deployment as similar to local dev as possible? Also adding the new environment variable to the template.env and docker-compose to pass it into django?

I added a rabbitmq service to docker-compose, and also updated the template.env.

PR #188, opened 2023-05-03
Use `django-configurations` for settings management, setup S3/MinIO

mvandenburgh:


django/src/rdwatch/server/settings.py
```diff
@@ -1,63 +1,125 @@
...
-from urllib.parse import urlparse
+import os
 
-SECRET_KEY = environ['RDWATCH_SECRET_KEY']
+from configurations import Configuration, values
 
+_environ_prefix = 'RDWATCH'
 
-ALLOWED_HOSTS = ['*']
-DEBUG = environ['RDWATCH_DJANGO_DEBUG'].lower() in ('1', 'true', 'yes', 'on')
```
BryonLewis, line 8: I think we need both of these (probably the DEBUG one modified) inside the DevelopmentConfiguration.  The server returns all 404's if the Allowed_Hosts isn't in there.  I assume probably in DevelopmentConfiguration we would want DEBUG set to one already or we can check for the environment Prefix 'RDWATCH_DJANGO'.

BryonLewis (changes requested):
There are some changes needed with the Environment variables to get it to work locally running with a simple docker compose up.

django/src/rdwatch/server/settings.py
mvandenburgh: Thanks, I think this is why I was having issues in our debug session yesterday.

BryonLewis (approved):
Did another quick test and it appears to work, I'll work on getting it integrated with my PR's and getting it to pass CI when that happens.

PR #190, opened 2023-05-08
Make Django max upload size consistent with NGINX config

mvandenburgh:
Our NGINX config allows up to 128 MB for JSON POST bodies, but Django has a default limit of 2.5 MB.


BryonLewis (approved):
Absolutely love how Byte -> MB conversion works and comes out with these lovely numbers.  That extra 24 on 1024 makes things a pain.

PR #191, opened 2023-05-08
Fix score possibly being set to NULL

mvandenburgh:
According to the site/region model specs, if the `score` field is missing from a given model, then it's assumed to be 1.0. We weren't accounting for this in our site model ingest routine, so this PR fixes that.


BryonLewis (approved):
:+1:

PR #192, opened 2023-05-09
Simplify development environment

mvandenburgh:
Simplifies the development environment by
- Removing VSCode container integration
	- This seems to have been broken for a while, and was unused
- Removing production docker images from the `docker-compose.yaml`.
	- This was an artifact of the previous deployment, before we switched to AWS ECS/Fargate.
- Removing the `unit-reloader` docker image and relying on the Django dev server's native hot-reload capabilities
- Removing the client from docker-compose, and just allowing the developer to run `npm run dev` natively instead of running it behind an nginx server in docker.

In general, I tried to make things closer in line with the best practices for [girder 4](https://github.com/girder/cookiecutter-girder-4).

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -48,6 +48,12 @@ but allows developers to run Python code on their native system.
 5. To destroy the stack and start fresh, run `docker compose down`
    1. Note: this command does not destroy docker volumes, such as those associated with the postgresql and minio services. To destroy those as well, run `docker compose down -v`.
 
+### Running DB migrations manually
+The docker entrypoint script runs database migrations on container start. However, in some cases you may wish to run migrations manually;
+to do this, run the following command:
+
+`poetry --directory django run django-admin migrate`
+
 ## Type support for ".vue" imports in VS Code
 
 Enable ["takeover mode"](https://github.com/johnsoncodehk/volar/discussions/471) for Volar.


mvandenburgh:
Sorry, looks like this is failing when i run it locally... will rerequest review once fixed

dev/export-env.sh
```diff
@@ -0,0 +1,29 @@
+# Export environment variables from the .env file in the first argument.
```
mvandenburgh, line 1: This file + the rest of the files in this directory are based on https://github.com/girder/cookiecutter-girder-4/tree/master/%7B%7B%20cookiecutter.project_slug%20%7D%7D/dev

.github/workflows/cd.yaml
```diff
@@ -11,22 +11,7 @@ concurrency:
...
-            target:
-          - image-name: builder
-            context: "."
-            target: builder
-          - image-name: postgresql
-            context: docker/services/postgresql
-            target:
-          - image-name: unit-reloader
-            context: docker/services/debug/unit-reloader
-            target:
```
mvandenburgh, line 28: Now that we're not using these in development, there's no need to push them as separate images.

dev/.env.docker-compose
```diff
@@ -0,0 +1,13 @@
...
+RDWATCH_POSTGRESQL_URI="postgresql://rdwatch:secretkey@postgresql:5432/rdwatch"
+RDWATCH_REDIS_URI="redis://:secretkey@redis-master:6379"
+RDWATCH_CELERY_BROKER_URL="amqp://rdwatch:secretkey@rabbitmq:5672/"
+RDWATCH_DJANGO_DEBUG=1
+RDWATCH_ACCENTURE_VERSION="2"
+RDWATCH_MINIO_STORAGE_ENDPOINT=minio:9000
+RDWATCH_MINIO_STORAGE_ACCESS_KEY=rdwatch
+RDWATCH_MINIO_STORAGE_SECRET_KEY=secretkey
+RDWATCH_STORAGE_BUCKET_NAME=rdwatch
+AWS_REQUEST_PAYER=requester
```
BryonLewis, line 13: Some indications even if commented out of the Environment variables for AWS Access:
`AWS_ACCESS_KEY_ID`
`AWS_SECRET_ACCESS_KEY`

docker-compose.override.yaml
```diff
@@ -1,86 +1,58 @@
...
-      RDWATCH_MINIO_STORAGE_ENDPOINT: minio:9002
-      RDWATCH_STORAGE_BUCKET_NAME: rdwatch
-    profiles:
-      - debug
-      - tools
-      - vscode
+      AWS_ACCESS_KEY_ID: ${AWS_ACCESS_KEY_ID:?}
+      AWS_SECRET_ACCESS_KEY: ${AWS_SECRET_ACCESS_KEY:?}
+      RDWATCH_SMART_STAC_URL: ${RDWATCH_SMART_STAC_URL:?}
+      RDWATCH_SMART_STAC_KEY: ${RDWATCH_SMART_STAC_KEY:?}
```
BryonLewis, line 50: I think we need the broker URL in here
`RDWATCH_CELERY_BROKER_URL: ${RDWATCH_CELERY_BROKER_URL:?}`

dev/.env.docker-compose
BryonLewis, line 13: Additionally for local testing probably some example of the environment variables:
`      RDWATCH_SMART_STAC_URL: ${RDWATCH_SMART_STAC_URL:?}
      RDWATCH_SMART_STAC_KEY: ${RDWATCH_SMART_STAC_KEY:?}
`

DEVELOPMENT.md
```diff
@@ -2,39 +2,45 @@
...
+### Run Application
+1.  Ensure `docker compose -f ./docker-compose.yaml up -d` is still active
+2. Run:
+   1. `source ./dev/export-env.sh`
+   2. `poetry run --directory django django-admin runserver`
+3. Run in a separate terminal:
+   1. `source ./dev/export-env.sh`
+   2. `poetry run --directory django celery --app rdwatch.celery worker --loglevel INFO --without-heartbeat`
+4. When finished, run `docker compose stop`
+5. To destroy the stack and start fresh, run `docker compose down -v`
```
BryonLewis, line 41: `-v` will remove/destroy any volumes.  Not sure if you want to destroy the `postgresql-data` and `minio-data` volumes?

BryonLewis (changes requested):
What are your thoughts about adding a .gitignore and ..env.template.private to the dev folder which contains a stub of the private/secret keys we want to use but don't want accidently committed to github?

We should be able to specify multiple environment variables (https://docs.docker.com/compose/environment-variables/set-environment-variables/#use-the-env_file-attribute)

That way we have the original .env.docker-compose and then a .env.private and we should be able to have a (mostly minus the UI) fully functioning system with a single docker-compose up.


I.E: 
`RDWATCH_SMART_STAC_KEY`
`AWS_ACCESS_KEY_ID`
`AWS_SECRET_ACCESS_KEY`


I'm a bit torn about having a fully contained system running with the client side when using docker-compose up.  It makes it easier for others to get something launched and running with a UI.  I do like doing my client dev independently (launching it outside of a container).  For now I think it's fine but we may want to revisit i the future and have some version of the UI launch (not on port 8080, but somewhere with a simple docker compose up)

docker-compose.override.yaml
mvandenburgh: This is added in the `.env.docker-compose` and `.env.docker-compose-native` files.

DEVELOPMENT.md
mvandenburgh: Added additional instructions clarifying this https://github.com/ResonantGeoData/RD-WATCH/pull/192/commits/3ae6ee62b52631e8ffd99ea8f7392d29b151b87d

DEVELOPMENT.md
```diff
@@ -2,39 +2,45 @@
...
-## Quick start
-
-Get up and running with the recommended development setup. It's advised to develop in the provided container since we depend on some system dependencies (e.g. GDAL).
-
-1. Make a copy of the file ["template.env"](https://github.com/ResonantGeoData/RD-WATCH/blob/phase-ii/template.env) as ".env" and fill it out
-2. Pull in the Docker images: `docker compose --profile vscode pull`
-3. Install the dependencies for Django: `docker compose run --rm poetry install --all-extras`
-4. Install the dependencies for Vue: `docker compose run --rm npm install`
-5. Run migrations: `docker compose run --rm poetry run django-admin migrate`
-6. Load test data: `docker compose run --rm poetry run django-admin loaddata testdata`
```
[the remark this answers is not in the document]
mvandenburgh: I think it may be better to have a [Django management command](https://docs.djangoproject.com/en/4.2/howto/custom-management-commands/) that creates sample data using the ORM. This would eliminate the need to update the sample data whenever a new migration is introduced.

.github/workflows/cd.yaml
```diff
@@ -52,11 +37,10 @@ jobs:
         with:
           cache-from: type=gha
           cache-to: type=gha,mode=max
-          context: ${{ matrix.context }}
+          context: .
           platforms: linux/amd64,linux/arm64
```
mvandenburgh, line 41: I would also love to remove the build step for `linux/arm64` as it seems to make the build CI job take longer than it needs to, but I'll leave it in place for now since I'm not completely sure that it's not used anywhere. Alternatively, I'd like to see if we can parallelize the builds.

dev/.env.docker-compose
mvandenburgh: Added .env.private support https://github.com/ResonantGeoData/RD-WATCH/pull/192/commits/cc31451b38dc0cfaab362e35181ba8de17b68419

DEVELOPMENT.md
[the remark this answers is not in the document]
BryonLewis: Agreed that would be much better.  Although we want some reasonable locations/polys and states tested so it would need to be curated sample data.

BryonLewis (changes requested):
I think this may require a bigger/different change.  If you completely remove the root environment file the docker compose doesn't know what to see the values to.   I believe the `env_file` option in the docker compose sets the environment file for that container, but docker compose doesn't know about that environment file. So without a root .env file in the project docker compose doesn't know about all of the environment settings we have set.

Example would be the PostgreSQL service:
In the docker compose we have:
`POSTGRES_PASSWORD:  ${RDWATCH_SECRET_KEY:?}`
If you don't have a root `.env` file this won't work.
Even if you pass it the `env_file: ./dev/.env.docker-compose` it won't work because that sets the container's environment value so internal to the container RDWATCH_SECRET_KEY is a value but POSTGRES_PASSWORD isn't.

I believe you may want to test this with your system environment tool turned off and a .env removed as well.

mvandenburgh:
> I think this may require a bigger/different change. If you completely remove the root environment file the docker compose doesn't know what to see the values to. I believe the `env_file` option in the docker compose sets the environment file for that container, but docker compose doesn't know about that environment file. So without a root .env file in the project docker compose doesn't know about all of the environment settings we have set.
> 
> Example would be the PostgreSQL service: In the docker compose we have: `POSTGRES_PASSWORD: ${RDWATCH_SECRET_KEY:?}` If you don't have a root `.env` file this won't work. Even if you pass it the `env_file: ./dev/.env.docker-compose` it won't work because that sets the container's environment value so internal to the container RDWATCH_SECRET_KEY is a value but POSTGRES_PASSWORD isn't.
> 
> I believe you may want to test this with your system environment tool turned off and a .env removed as well.

Yeah sorry i thought i disabled the env tool i use when testing this. I'll do a fresh clone of this branch in a clean directory and do a compose up and see if I can reproduce this.

BryonLewis:
No problem.  You may run into some additional issues with a fresh clone as well.  The django-admin isn't installed by default in the default build.  (I don't know if we want to make a build context where it installs it, maybe?) It is relying on you running poetry install before ever bringing up the system.  If you docker compose build/up before poetry install it will locked to the user of the container so you need to sudo rm -rf .venv to install locally then.

mvandenburgh:
@BryonLewis I resolved the env var issues, this should work now. can you try running docker compose using the new instructions?

BryonLewis:
Had a brief discussion on the `.venv` issue and it being used by the container and locally.

One option would be to create a /poetry/venvs directory in the container `RUN mkdir -p /poetry/venvs`
swapping the in-project to false and setting the virutalenvs.path to the new directory:
` && poetry config virtualenvs.in-project false \
 && poetry config virtualenvs.path /poetry/venvs
 `
 
 Changing the `.dockerignore` file to include the `django/.venvs` directory so it does copy over the local version in the context or when mounting.

Finally I think there were some missing environment variables from the `./dev/.env.docker-compose` files for the REDIS URL and CELERY broker URL.  I wasn't able to run the poetry run manage.py commands outside of docker because it didn't know what the values should be in context.  I would also check that the proper ports are exposed to the host-pc for the containers that would be needed if running manage.py outside the container.

BryonLewis (changes requested):
Most of it works much better just one thing (minio-storage).  I think the only thing left is to somehow get django-minio installed in celery/django when using the docker-compose.override.yaml.

DEVELOPMENT.md
```diff
@@ -2,39 +2,53 @@
...
+2. Set the environment variables in `.env`.
+3. Run `docker compose -f ./docker-compose.yaml up -d`
+4. Install Python 3.10
+5. Install
+   [`psycopg2` build prerequisites](https://www.psycopg.org/docs/install.html#build-prerequisites)
+6. Install Poetry
+7. Run `poetry --directory django install`
+8. Run the following commands to configure your environment:
+   1. `source ./dev/export-env.sh dev/.env.docker-compose-native`
+   2. `source ./dev/export-env.sh .env`
```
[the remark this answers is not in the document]
mvandenburgh: :+1: https://github.com/ResonantGeoData/RD-WATCH/pull/192/commits/c6629f001c2cf0e21402c5e677a6aeb381b9d6fc

mvandenburgh:
> Most of it works much better just one thing (minio-storage). I think the only thing left is to somehow get django-minio installed in celery/django when using the docker-compose.override.yaml.

I added a new build stage that installs dev dependencies and pointed celery/django services in docker-compose.override.yaml at it - https://github.com/ResonantGeoData/RD-WATCH/commit/7b18e4d9dc19833c932ac65656ecab8885e1aeb2

BryonLewis (approved):
Was able to do a docker compose up after creating my proper .env file and the system seemed to come up properly (I needed to do a merge of main to make the /region endpoint work).

Then I ran `docker compose run --rm django poetry run django-admin migrate`
Followed by `docker compose run --rm django poetry run django-admin loaddata testdata` to get test data loading and running.

Ran a download of WV images and it seems to work.

The only other minor suggestion is adding the migra/loaddata instructions to the DEVELOPMENT.md (`docker compose run --rm django poetry run django-admin loaddata testdata)` to let people know how to get data into the system.

PR #193, opened 2023-05-09
Refactor Site Observation Image Fetching

BryonLewis:
Based on the comments on the previous PR I've refactored and improved a couple things.

- Only one function `get_range_captures` where closest capture just specifies a timebuffer of 1 day and then filters for the closest capture
- The Temporary File writing is eliminated, I could write the bytes directly into a Django File object instead of requiring any temporary file.
- changed from `get-images` to `generate-images` in frontend/backend
- Cleaned up some dangling stuff for area calculations in the client.

### DIFF
--- a/django/src/rdwatch/tasks/__init__.py
+++ b/django/src/rdwatch/tasks/__init__.py
@@ -219,7 +219,7 @@ def get_siteobservations_images(
                 timestamp=capture.timestamp,
                 source=baseConstellation,
             )
-            if found.count() > 0:
+            if found.exists():
                 existing = found.first()
                 existing.image.delete()
                 existing.cloudcover = cloudcover
--- a/django/src/rdwatch/tasks/__init__.py
+++ b/django/src/rdwatch/tasks/__init__.py
@@ -1,6 +1,5 @@
 import io
 import logging
-import os
 from datetime import datetime, timedelta
 
 from celery import shared_task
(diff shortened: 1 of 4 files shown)


django/src/rdwatch/tasks/__init__.py
```diff
@@ -249,27 +213,23 @@ def get_siteobservations_images(
...
-            with open(output, 'rb') as imageFile:
-                image = File(imageFile, output)
-                found = SiteImage.objects.filter(
+            image = File(io.BytesIO(bytes), name=output)
+            found = SiteImage.objects.filter(
+                siteeval=baseSiteEval,
+                timestamp=capture.timestamp,
+                source=baseConstellation,
+            )
+            if found.count() > 0:
```
mvandenburgh, line 222:
```suggestion
            if found.exists():
```
If we just want to check for existence, this should generate a more efficient query than an expensive `COUNT`. https://docs.djangoproject.com/en/4.2/ref/models/querysets/#django.db.models.query.QuerySet.exists

django/src/rdwatch/tasks/__init__.py
```diff
@@ -196,26 +164,22 @@ def get_siteobservations_images(
...
-                        siteeval=observation.siteeval,
-                        siteobs=observation,
-                        timestamp=observation.timestamp,
-                        image=image,
-                        cloudcover=cloudcover,
-                        source=baseConstellation,
-                    )
-            os.remove(output)
+            image = File(io.BytesIO(bytes), name=output)
+            if found.count() > 0:
```
mvandenburgh, line 168: See below comment.
```suggestion
            if found.exists():
```

mvandenburgh (approved):
Found one more possible optimization, but otherwise this LGTM :+1:

PR #195, opened 2023-05-11
Running Tasks

BryonLewis:
Backend
- Add SatelliteFetching table to record status of retrieving satellite images (Running, Complete, Error)
    - Status will initiate to Running and end with Complete
    - Added field for eventual Errors
    - Open to any suggestions for a better way to do this
- In main data for siteEvaluation return the SatelliteFetching status if it exists under the 'job' category
- Add PercentBlack to the SiteImage Table so we can filter on percentage black/nodata images.
- Updated S3/WV image range scaling to utilize 0-10000 instead of the previous scaling ranges

Frontend:
- Fix site Image toggling issues
- Implement polling when 'Running' state is found every X seconds and indicate that images are being downloaded in the UI
- Add Settings/Filters for the SiteImage display
    - Filter based on cloud cover of the images displayed
    - Filter based on percentage black (default it to some reasonable value like 70%)
    - Opacity slider for all of these images to adjust the image opacity (default to 100%)
- Add scrolling to the selection list for selected Observations.
- Place the new filters into some settings system for the site Evaluations section.  (Clean up the S2/WV/Clear all settings)


mvandenburgh:
@BryonLewis I attempted to fix some of the race conditions/concurrency issues we discussed earlier. I thought it was easier to just push a separate branch than leave multiple github code comments - https://github.com/ResonantGeoData/RD-WATCH/pull/200

mvandenburgh (changes requested):

django/src/rdwatch/views/tile.py
```diff
@@ -285,7 +285,7 @@ def satelliteimage_raster_bbox(
     return get_satelliteimage_raster(request)
 
 
-@cache_page(60 * 60 * 24 * 365)
+# @cache_page(60 * 60 * 24 * 365)
```
mvandenburgh, line 288:
```suggestion
@cache_page(60 * 60 * 24 * 365)
```

mvandenburgh (approved):

PR #197, opened 2023-05-13
Remove dead/unused code

mvandenburgh:
Based on #192 and should be merged after that PR.

- Removes a `dataclass` that is no longer used
- Remove the server status endpoint - this isn't used anywhere.


BryonLewis (changes requested):
I'm using the server status to check to see that the container is healthy so I can start the celery container.  We either need to remove the migration from the entrypoint (which may be a good idea to prevent bad migrations).  Or find another way to prevent both the celery and django container in the dev environment from attempting to run migrations in parallel (something that can cause issues)

mvandenburgh:
> I'm using the server status to check to see that the container is healthy so I can start the celery container. We either need to remove the migration from the entrypoint (which may be a good idea to prevent bad migrations). Or find another way to prevent both the celery and django container in the dev environment from attempting to run migrations in parallel (something that can cause issues)

ok. For now I dropped that change, but I agree we should find a way to handle migrations better

django/src/rdwatch/dataclasses/server_status.py
```diff
@@ -1,9 +0,0 @@
-from dataclasses import dataclass
-from datetime import timedelta
-
-
-@dataclass
-class ServerStatus:
-    uptime: timedelta
-    hostname: str
-    ip: str
```
BryonLewis, line 9: I think that ./rdwatch/serializers/server_status.py needs this as an import?

BryonLewis (changes requested):

BryonLewis (approved):
pulled, tested, and LGTM. :+1:

PR #198, opened 2023-05-16
Migrate to new dev-dependency format for poetry

mvandenburgh:
The current format we use is deprecated and will be removed in future versions of poetry
https://python-poetry.org/docs/managing-dependencies/#dependency-groups (see the section "A note about defining a dev dependencies group")


BryonLewis (approved):
LGTM

PR #200, opened 2023-05-16
Fix race conditions in image gen task

mvandenburgh:


mvandenburgh (commented):
This doesn't address the issue of a task dying and leaving itself stuck in a "RUNNING" state, but it at least avoids clobbering/overwriting of data.

BryonLewis (approved):
Pulled, tested correctly returned that task was already running as well.  Thanks.

PR #201, opened 2023-05-17
Set timestamp when inserting SatelliteFetching row

BryonLewis:


BryonLewis:
I was going to fix this but saw you already had the branch, I pulled and tested locally and it fixed the issue with new downloads.  Me creating the PR is approval so you can merge when you feel ready to deploy to the server.

mvandenburgh (approved):

PR #202, opened 2023-05-17
Various Fixes

BryonLewis:
- The Status wasn't included when initializing a sitefetching table
- Added some error handling for a URLError I saw during some testing
- Somehow empty images were getting created with null image and percent_black being null.  While they can initially have this it shouldn't be something that exists.  I added a check in the task logic to skip over images that return None and on the client side it will filter any existing data which has null.
-

### DIFF
--- a/django/src/rdwatch/tasks/__init__.py
+++ b/django/src/rdwatch/tasks/__init__.py
@@ -1,6 +1,7 @@
 import io
 import logging
 from datetime import datetime, timedelta
+from urllib.error import URLError
 
 from celery import shared_task
 from PIL import Image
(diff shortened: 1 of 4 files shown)


django/src/rdwatch/tasks/__init__.py
```diff
@@ -96,7 +97,15 @@ def fetch_boundbox_image(
...
     timebuffer = timedelta(days=1)
-    captures = get_range_captures(bbox, timestamp, constellation, timebuffer, worldView)
+    try:
+        captures = get_range_captures(
+            bbox, timestamp, constellation, timebuffer, worldView
+        )
+    except URLError as e:
+        logger.warning('Failed to get range capture because of URLError')
+        logger.warning(e)
+        return None
```
mvandenburgh, line 107: What circumstances would cause a URLError? And should we be setting the task status to "ERROR" accordingly?

mvandenburgh (approved):

django/src/rdwatch/tasks/__init__.py
BryonLewis: I'm not entirely sure, it's not a reproducible thing, it was happening last night so I wanted to account for it.  I may be tempted to through a full catch in there and log error for anything that may happen between searching/downloading S3 images.

BryonLewis: agreed

PR #204, opened 2023-05-19
Convert region endpoints to Ninja

mvandenburgh:


django/src/rdwatch/views/region.py
```diff
@@ -1,13 +1,39 @@
...
-from rest_framework.pagination import PageNumberPagination
+import iso3166
+from ninja import Schema
+from ninja.pagination import RouterPaginated
+
+from django.http import HttpRequest
+from django.shortcuts import get_object_or_404
 
 from rdwatch.models import Region
-from rdwatch.serializers import RegionSerializer
```
mvandenburgh, line 5: I didn't remove this serializer yet since it's used elsewhere. But eventually, `RegionSerializer` will be replaced by the new `RegionSchema`

vue/src/client/models/RegionList.ts
```diff
@@ -4,5 +4,5 @@ export type RegionList = {
   count: number;
   next: null;
   previous: null;
```
BryonLewis, line 6: are these still valid with the new pagination?  I forgot on the last PR to check the modelruns.  I know we don't ever have more than one page, but there are no null/string values in the response for this.  Probably should update performers as well (I didn't check that last time).

django/src/rdwatch/views/region.py
```diff
@@ -1,13 +1,39 @@
...
+        classification: str = obj.classification.slug
+        number = obj.number
+
+        country_numeric = str(country).zfill(3)
+        country_code = iso3166.countries_by_numeric[country_numeric].alpha2
+        region_number = 'xxx' if number is None else str(number).zfill(3)
+        return f'{country_code}_{classification}{region_number}'
+
+
+router = RouterPaginated()
```
BryonLewis, line 27: wasn't an issue with performers but we have ~60 regions so the limit on pages for regions by default I think is 100.  Just so we know if we start hitting a high number of regions.

BryonLewis (approved):
Just a few comments about the default pagination size and if we use this for other items to be wary of it.

Also modification to the type responses on the client to remove the next/prev because I don't think they're needed anymore.

PR #205, opened 2023-05-25
Swapping to Vuetify for Component Library

BryonLewis:
Started attempting to convert us over to to vuetify for the component library instead of using daisyUI and making our own components.  This will make it easier to do some changes as well as color tools that can be used to specify colors or add colors to different components.

Done:

- Vuetify imported and loaded in
- SideBar.vue, ModelRunList, ModelRunDetails, Settings, and the Performer/Region filters have been converted.
- Removed the SiteEvaluation.vue and SiteEvaluationDetails.vue because they didn't seem to be used.
- Remove DaisyUI but kept tailwind if it is used.
- The custom created filters for the Region/Performer dropdown were replaced with clearable v-selects.
- Colors are slightly different but I can work on a theme to correct this if needed.
- The Tooltip is now a true Vue Component instead of HTML with TailwindCSS, this will allow more customization in the future.

image

Mostly just check that stuff behaves properly and give feedback where things don't look right.

Next up may be re-doing the logic for requests.  Not a big fan of the watchEffect causing reloads in a complicated function.


vue/package.json
```diff
@@ -22,29 +22,32 @@
     "@heroicons/vue": "^2.0.13",
```
mvandenburgh, line 22: This can also be removed, I think, if we are completed converted over to material design icons.

PR #206, opened 2023-05-31
Celery-Results and Cancel Tasks

BryonLewis:
Installs and adds django-celery-results and enables the capability to cancel tasks.

- The satellite fetching will now store the celery Id for the running task. (Remember we only allow one running task per site observation).
- If the satellite fetching table is running you can cancel the task.  It will attempt to load the celery tasks and terminate it.  It will also reset the state of the satellite fetching to Complete state so another task can be run.
- I added some basic progress for the different sections of the task (I didn't connect them up to the client because I'm not sure if it is necessary yet).
- Celery_Time_limit has ben set to 1200 seconds (20 minutes): [ CELERYD_TIME_LIMIT](https://docs.celeryq.dev/en/stable/userguide/configuration.html#task-time-limit)
- Followed Girder Worker Plugic job cancellation with the PUT method.  Let me know if this interferes with the deployment (I know we had an issue with using POST because of the ALB or something like that).
- Detailed Info about the Job:
   - Starts with attempting to get the Site Observations and provides the X of Y status
   - Then does a search for Image Captures
   - Then Downloads Image Captures again doing X of Y status.

image

NOTE:  After merging to main with the dev environment change I noticed that on first launch there would be an issue with migrations specifically in the celery server.  I believe this was an issue with two containers attempting to run migrations at the same time through the docker entrypoint.  We may want to remove that out of the entry point to prevent this.  For now I made it so the celery container won't start until after the django container is fully up and responsive (This should prevent migration issues of it being run in parallel).


django/src/rdwatch/views/site_observation.py
```diff
@@ -97,10 +99,18 @@ def site_observations(request: HttpRequest, pk: int):
     output['images'] = image_serializer.data
     if SatelliteFetching.objects.filter(siteeval=pk).exists():
         retrieved = SatelliteFetching.objects.filter(siteeval=pk).first()
+        celery_data = {}
+        if retrieved.celery_id != '':
```
mvandenburgh, line 103: The "pythonic" way to check for empty strings is
```suggestion
        if retrieved.celery_id:
```

docker-compose.override.yaml
```diff
@@ -49,8 +55,13 @@ services:
...
+      django:
+        condition: service_healthy
+      postgresql:
+        condition: service_started
+      minio:
+        condition: service_started
+      redis-master:
+        condition: service_started
+      rabbitmq:
+        condition: service_started
```
mvandenburgh, line 67: These will have to be rebased on the changes from #192.

mvandenburgh (approved):
Looks good! One minor nitpick about conditionals

django/src/rdwatch/views/site_observation.py
BryonLewis: Yeah, I do this a lot even in JS, mostly because I like being explicit about things and not relying on an implicit conversion in your head when looking through the code.

docker-compose.override.yaml
BryonLewis: I thought I already merged those changes in this morning before I did this?

PR #208, opened 2023-06-01
Support missing `score` for site observation

mvandenburgh:
#191 fixed this for site evaluations, but not site observations; this PR fixes that.


mvandenburgh:
@BryonLewis I'm going to merge this to unblock t&e, but feel free to still review if you'd like.

PR #210, opened 2023-06-05
Don't run poetry install in docker entrypoint

mvandenburgh:
This is already done in the Dockerfile and shouldn't need to be repeated. It also seems to be causing a deployment failure when I tried deploying the latest image, logs are here https://us-west-2.console.aws.amazon.com/ecs/v2/clusters/resonantgeodatablue-fargate-cluster/services/resonantgeodatablue-worker-service/tasks/762716d1bb7a43e7a3fdd18d5b23f873/logs?region=us-west-2

### DIFF
--- a/docker/docker-entrypoint.sh
+++ b/docker/docker-entrypoint.sh
@@ -21,7 +21,6 @@ if [ "$1" = "unitd" ]; then
   fi
 fi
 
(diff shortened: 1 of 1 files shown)


BryonLewis (approved):
correct in that update you moved it into the docker-file so it shouldn't need to be here.

PR #211, opened 2023-06-05
Build docker images in parallel

mvandenburgh:
This should help speed up builds, or at least prevent the amd64 image build from getting held up by the arm64 one.

See https://docs.docker.com/build/ci/github-actions/multi-platform/#distribute-build-across-multiple-runners


mvandenburgh:
Oops, i did this wrong i think....

PR #213, opened 2023-06-07
Update nginx-unit venv path

mvandenburgh:
Follow up to #212, this was also causing deployment failures.


mvandenburgh:
Docs: https://unit.nginx.org/configuration/

PR #214, opened 2023-06-07
RDWATCH Scoring

BryonLewis:
Adding in Scoring capabilities.

### DIFF
--- a/django/src/rdwatch/api.py
+++ b/django/src/rdwatch/api.py
@@ -1,5 +1,7 @@
 from ninja import NinjaAPI
 
+from rdwatch_scoring.views import router as scores
+
 from .views.performer import router as performer_router
 from .views.region import router as region_router
 from .views.site_model import router as site_model_router
@@ -9,3 +11,5 @@ api = NinjaAPI()
 api.add_router('/performers/', performer_router)
 api.add_router('/regions/', region_router)
 api.add_router('/model-runs/', site_model_router)
+# Note lets not use this, try to implement it in the rdwatch_scoring app
+api.add_router('/scores/', scores)
--- a/django/src/rdwatch/server/settings.py
+++ b/django/src/rdwatch/server/settings.py
@@ -31,6 +31,8 @@ class BaseConfiguration(Configuration):
         'django_extensions',
         'rdwatch',
         'django_celery_results',
+        # Maybe conditionally add Scoring
+        'rdwatch_scoring',
     ]
 
     MIDDLEWARE = [
@@ -63,14 +65,25 @@ class BaseConfiguration(Configuration):
         environ_prefix=_environ_prefix,
     )
 
-    DATABASES = values.DatabaseURLValue(
+    DB_val = values.DatabaseURLValue(
+        alias='default',
         environ_name='POSTGRESQL_URI',
         environ_prefix=_environ_prefix,
         environ_required=True,
         # Additional kwargs to DatabaseURLValue are passed to dj-database-url
         engine='django.contrib.gis.db.backends.postgis',
-        conn_max_age=None,
     )
+    scoring_val = values.DatabaseURLValue(
+        'postgres://scoring:secretkey@scoredb:5432/scoring',
+        alias='scoringdb',
+        engine='django.contrib.gis.db.backends.postgis',
+    )
+    db_dict = DB_val.value
+    scoring_dict = scoring_val.value
+
+    # NOTE - CHANGE so an environment variable can add this in
+    db_dict.update(scoring_dict)
+    DATABASES = db_dict
 
     # Set to same value allowed by NGINX Unit server in `settings.http.max_body_size`
     # (in /docker/nginx.json)
@@ -92,6 +105,8 @@ class BaseConfiguration(Configuration):
     CELERY_CACHE_BACKEND = 'django-cache'
     CELERY_RESULT_EXTENDED = True
     CELERYD_TIME_LIMIT = 1200
+    # Make this conditional based on environment variable as well
+    DATABASE_ROUTERS = ['rdwatch_scoring.router.ScoringRouter']
 
 
 class DevelopmentConfiguration(BaseConfiguration):
--- a/django/src/rdwatch_scoring/api.py
+++ b/django/src/rdwatch_scoring/api.py
@@ -1,0 +1,7 @@
+from ninja import NinjaAPI
+
+from .views import router as scores
+
+api = NinjaAPI()
+
+api.add_router('/scores', scores)
--- a/django/src/rdwatch_scoring/router.py
+++ b/django/src/rdwatch_scoring/router.py
@@ -1,0 +1,40 @@
+class ScoringRouter:
+    """
+    A router to control all database operations on models in the
+    auth and contenttypes applications.
+    """
+
+    route_app_labels = {'rdwatch_scoring'}
+
+    def db_for_read(self, model, **hints):
+        """
+        Attempts to read scoring and contenttypes models go to scoringdb.
+        """
+        if model._meta.app_label in self.route_app_labels:
+            return 'scoringdb'
+        return None
+
+    def db_for_write(self, model, **hints):
+        """
+        Attempts to write scoringdb models go to auth_db.
+        """
+        if model._meta.app_label in self.route_app_labels:
+            return 'scoringdb'
+        return None
+
+    def allow_relation(self, obj1, obj2, **hints):
+        """
+        Allow relations if a model in the scorin apps is
+        involved.
+        """
+        if (
+            obj1._meta.app_label in self.route_app_labels
+            or obj2._meta.app_label in self.route_app_labels
+        ):
+            return True
+        return None
+
+    def allow_migrate(self, db, app_label, model_name=None, **hints):
+        if app_label in self.route_app_labels:
+            return db == 'scoringdb'
+        return None
--- a/django/src/rdwatch_scoring/urls.py
+++ b/django/src/rdwatch_scoring/urls.py
@@ -1,0 +1,12 @@
+from django.urls import path
+from rest_framework import routers
+
+from rdwatch_scoring.api import api
+
+router = routers.SimpleRouter(trailing_slash=False)
+
+urlpatterns = [
+    path('/api/scoring', api.urls),
+]
+
+urlpatterns += router.urls
--- a/django/src/rdwatch_scoring/views.py
+++ b/django/src/rdwatch_scoring/views.py
@@ -1,0 +1,30 @@
+from django.shortcuts import render
+
+# Create your views here.
+from django.http import HttpRequest
+from django.shortcuts import get_object_or_404
+from ninja import Schema
+from ninja.pagination import RouterPaginated
+import logging
+
+from .models import Observation
+logger = logging.getLogger(__name__)
+
+
+class ScoreObservationSchema(Schema):
+    uuid: str = None
+    name: str = None
+    source: str = None
+    sensor: str = None
+    phase: str = None
+    score: str = None
+
+router = RouterPaginated()
+
+
+@router.get('/', response=list[ScoreObservationSchema])
+def list_regions(request: HttpRequest):
+
+    objects = Observation.objects.all()
+    logger.warning(objects)
+    return objects
--- a/docker-compose.override.yaml
+++ b/docker-compose.override.yaml
@@ -65,3 +65,17 @@ services:
         condition: service_started
       django:
         condition: service_healthy
+  # scoring database
+  scoredb:
+    image: postgis/postgis:latest
+    environment:
+      POSTGRES_USER: scoring
+      POSTGRES_DB: scoring
+      POSTGRES_PASSWORD: secretkey
+    volumes:
+      - postgresql-scoring-data:/var/lib/postgresql/data
+    ports:
+      - 5433:5432
+
+volumes:
+  postgresql-scoring-data:
--- a/django/pyproject.toml
+++ b/django/pyproject.toml
@@ -19,7 +19,10 @@ classifiers = [
   "Framework :: Django :: 4.0",
   "Framework :: Django",
 ]
-packages = [ { include = "rdwatch", from = "src" } ]
+packages = [
+  { include = "rdwatch", from = "src" },
+  { include = "rdwatch_scoring", from = "src/rdwatch_scoring" }
+]
 
 [tool.poetry.dependencies]
 python = "~3.10.0"
--- a/django/src/rdwatch_scoring/admin.py
+++ b/django/src/rdwatch_scoring/admin.py
@@ -1,0 +1,1 @@
+# Register your models here.
--- a/django/src/rdwatch_scoring/apps.py
+++ b/django/src/rdwatch_scoring/apps.py
@@ -1,0 +1,6 @@
+from django.apps import AppConfig
+
+
+class ScoringConfig(AppConfig):
+    default_auto_field = 'django.db.models.BigAutoField'
+    name = 'rdwatch_scoring'
--- a/django/src/rdwatch_scoring/models.py
+++ b/django/src/rdwatch_scoring/models.py
@@ -1,0 +1,1009 @@
+from django.contrib.gis.db import models
+
+
+class Addr(models.Model):
+    gid = models.AutoField(primary_key=True)
+    tlid = models.BigIntegerField(blank=True, null=True)
+    fromhn = models.CharField(max_length=12, blank=True, null=True)
+    tohn = models.CharField(max_length=12, blank=True, null=True)
+    side = models.CharField(max_length=1, blank=True, null=True)
+    zip = models.CharField(max_length=5, blank=True, null=True)
+    plus4 = models.CharField(max_length=4, blank=True, null=True)
+    fromtyp = models.CharField(max_length=1, blank=True, null=True)
+    totyp = models.CharField(max_length=1, blank=True, null=True)
+    fromarmid = models.IntegerField(blank=True, null=True)
+    toarmid = models.IntegerField(blank=True, null=True)
+    arid = models.CharField(max_length=22, blank=True, null=True)
+    mtfcc = models.CharField(max_length=5, blank=True, null=True)
+    statefp = models.CharField(max_length=2, blank=True, null=True)
+
+    class Meta:
+        managed = False
+        db_table = 'addr'
+
+
+class Addrfeat(models.Model):
+    gid = models.AutoField(primary_key=True)
+    tlid = models.BigIntegerField(blank=True, null=True)
+    statefp = models.CharField(max_length=2)
+    aridl = models.CharField(max_length=22, blank=True, null=True)
+    aridr = models.CharField(max_length=22, blank=True, null=True)
+    linearid = models.CharField(max_length=22, blank=True, null=True)
+    fullname = models.CharField(max_length=100, blank=True, null=True)
+    lfromhn = models.CharField(max_length=12, blank=True, null=True)
+    ltohn = models.CharField(max_length=12, blank=True, null=True)
+    rfromhn = models.CharField(max_length=12, blank=True, null=True)
+    rtohn = models.CharField(max_length=12, blank=True, null=True)
+    zipl = models.CharField(max_length=5, blank=True, null=True)
+    zipr = models.CharField(max_length=5, blank=True, null=True)
+    edge_mtfcc = models.CharField(max_length=5, blank=True, null=True)
+    parityl = models.CharField(max_length=1, blank=True, null=True)
+    parityr = models.CharField(max_length=1, blank=True, null=True)
+    plus4l = models.CharField(max_length=4, blank=True, null=True)
+    plus4r = models.CharField(max_length=4, blank=True, null=True)
+    lfromtyp = models.CharField(max_length=1, blank=True, null=True)
+    ltotyp = models.CharField(max_length=1, blank=True, null=True)
+    rfromtyp = models.CharField(max_length=1, blank=True, null=True)
+    rtotyp = models.CharField(max_length=1, blank=True, null=True)
+    offsetl = models.CharField(max_length=1, blank=True, null=True)
+    offsetr = models.CharField(max_length=1, blank=True, null=True)
+    the_geom = models.LineStringField(srid=4269, blank=True, null=True)
+
+    class Meta:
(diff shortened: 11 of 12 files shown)


django/src/rdwatch/server/settings.py
```diff
@@ -64,14 +66,28 @@ class BaseConfiguration(Configuration):
...
+        environ_prefix=_environ_prefix,
+        environ_required=True,
+        # Additional kwargs to DatabaseURLValue are passed to dj-database-url
+        engine='django.contrib.gis.db.backends.postgis',
+    )
+    db_dict = DB_val.value
+    scoring_dict = scoring_val.value
+
+    db_dict.update(scoring_dict)
+    DATABASES = db_dict
```
mvandenburgh, line 90: I think we can make `DATABASES` a `property` and move this logic into there.

django/src/rdwatch/server/settings.py
```diff
@@ -93,6 +109,8 @@ class BaseConfiguration(Configuration):
     CELERY_CACHE_BACKEND = 'django-cache'
     CELERY_RESULT_EXTENDED = True
     CELERYD_TIME_LIMIT = 1200
+    # Make this conditional based on environment variable as well
+    DATABASE_ROUTERS = ['rdwatch_scoring.router.ScoringRouter']
```
mvandenburgh, line 113: I guess we would use a `property` here as well.

django/src/rdwatch_scoring/api.py
```diff
@@ -0,0 +1,7 @@
+from ninja import NinjaAPI
+
+from .views import router as scores
+
+api = NinjaAPI()
+
+api.add_router('/scores', scores)
```
mvandenburgh, line 7: The convention we've used so far is to include the trailing slash on the router prefixes:
```suggestion
api.add_router('/scores/', scores)
```

django/src/rdwatch_scoring/migrations/0001_initial.py
```diff
@@ -0,0 +1,1536 @@
...
+
+import django.contrib.gis.db.models.fields
+import django.db.models.deletion
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    initial = True
+
+    dependencies = []
```
mvandenburgh, line 11: iirc it's best practice in Django to ensure your migrations always run in a deterministic order, so we can enforce that by doing
```suggestion
    dependencies = [
        ('rdwatch', '0010_hyperparameters_evaluation_and_more'),
    ]
```

django/src/rdwatch_scoring/urls.py
```diff
@@ -0,0 +1,12 @@
...
+
+from rdwatch_scoring.api import api
+
+router = routers.SimpleRouter(trailing_slash=False)
+
+urlpatterns = [
+    path('/api/scoring', api.urls),
+]
+
+urlpatterns += router.urls
```
mvandenburgh, line 12: DRF router is not needed as we're only using ninja in these endpoints.
```suggestion
from django.urls import path

from rdwatch_scoring.api import api

urlpatterns = [
    path('/api/scoring', api.urls),
]
```

django/src/rdwatch_scoring/views.py
```diff
@@ -0,0 +1,229 @@
...
+    color: str = None
+
+
+# If the combination of configurationId and regionId has scoring
+@router.get('/has_scores')
+def has_scores(request: HttpRequest):
+    if 'configurationId' not in request.GET or 'regionId' not in request.GET:
+        return HttpResponseBadRequest()
+    configId = request.GET['configurationId']
+    regionId = request.GET['regionId']
```
mvandenburgh, line 99: See https://django-ninja.rest-framework.com/guides/input/query-params/
```suggestion
# If the combination of configurationId and regionId has scoring
@router.get('/has_scores')
def has_scores(request: HttpRequest, configuration_id: int, region_id: int):
```

django/src/rdwatch_scoring/views.py
```diff
@@ -0,0 +1,229 @@
...
+    # from the hyper parameters we need the evaluation and evaluation_run Ids
+    configuration = HyperParameters.objects.filter(pk=configId).first()
```
mvandenburgh, line 101: If this returns `None`, the server will fail with a 500. We can use `get_object_or_404` to make this fail gracefully-
```suggestion
    configuration = get_object_or_404(HyperParameters, pk=config_id)
```

django/src/rdwatch_scoring/views.py
```diff
@@ -0,0 +1,229 @@
...
+    ).exists()
+
+
+# Region scoring Map
+@router.get('/region_scoring_colors')
+def region_color_map(request: HttpRequest):
+    if 'configurationId' not in request.GET or 'regionId' not in request.GET:
+        return HttpResponseBadRequest()
+    configId = request.GET['configurationId']
+    regionId = request.GET['regionId']
```
mvandenburgh, line 121: See above
```suggestion
def region_color_map(request: HttpRequest, configuration_id: int, region_id: int):
```

django/src/rdwatch_scoring/views.py
```diff
@@ -0,0 +1,229 @@
...
+        'configurationId' not in request.GET
+        or 'regionId' not in request.GET
+        or 'siteNumber' not in request.GET
+        or 'version' not in request.GET
+    ):
+        return HttpResponseBadRequest()
+    configId = request.GET['configurationId']
+    regionId = request.GET['regionId']
+    siteNumber = request.GET['siteNumber']
+    version = request.GET['version']
```
mvandenburgh, line 171: See above
```suggestion
def list_regions(request: HttpRequest, configuration_id: int, region_id: int, version: str):
```

django/src/rdwatch_scoring/views.py
```diff
@@ -0,0 +1,229 @@
...
+            'evaluationId': evaluationId,
+            'evaluation_run': evaluation_run,
+            'performer': performer_name,
+            'evaluation_uuid': evaluationUUID,
+            'unionArea': unionArea,
+            'statusAnnotated': status_annotated,
+            'temporalIOU': tempDict,
+            'color': sm_color,
+        }
+    return HttpResponse('Scoring not Found', status=404)
```
mvandenburgh, line 229: Instead of an explicit conditional and 404 `HttpResponse`, you can do
```suggestion
    evaluation = get_object_or_404(
        EvaluationRun,
        evaluation_run_number=evaluation_run,
        evaluation_number=evaluationId,
        region=region_name,
        performer=performer_name,
    )
    evaluationUUID = evaluation.uuid
    # Now we can get the SiteId information for this evaluationId
    generatedSiteId = (
        f'{region_name}_{siteNumber.zfill(4)}_{performer_name}_{version}'
    )
    logger.warning(generatedSiteId)
    siteObj = Site.objects.filter(
        site_id=generatedSiteId, evaluation_run_uuid=evaluationUUID
    ).first()
    unionArea = siteObj.union_area
    status_annotated = siteObj.status_annotated
    temporalIOU = EvaluationActivityClassificationTemporalIou.objects.filter(
        site_proposal=generatedSiteId,
        evaluation_run_uuid=evaluationUUID,
        activity_type='overall',
    ).first()
    tempDict = {
        'site_preparation': None,
        'active_construction': None,
        'post_construction': None,
    }
    if temporalIOU:
        tempDict = {
            'site_preparation': (temporalIOU.site_preparation),
            'active_construction': (temporalIOU.active_construction),
            'post_construction': (temporalIOU.post_construction),
        }
    # To understand the color associated with it you need
    # to grab the proposals these are the site_truth_matched
    # table from the evaluation_broad_area_search_proposal
    sm_color = get_site_scoring_color(evaluationUUID, 'overall', generatedSiteId)

    return {
        'region_name': region_name,
        'evaluationId': evaluationId,
        'evaluation_run': evaluation_run,
        'performer': performer_name,
        'evaluation_uuid': evaluationUUID,
        'unionArea': unionArea,
        'statusAnnotated': status_annotated,
        'temporalIOU': tempDict,
        'color': sm_color,
    }
```

django/tox.ini
```diff
@@ -40,8 +41,9 @@ passenv =
     RDWATCH_DJANGO_DEBUG
     RDWATCH_SMART_STAC_URL
     RDWATCH_SMART_STAC_KEY
+    RDWATCH_POSTGRESQL_SCORING_URI
 commands =
-    {envpython} src/manage.py makemigrations --check --dry-run
+    {envpython} src/manage.py makemigrations rdwatch --check --dry-run
```
mvandenburgh, line 46:
```suggestion
    {envpython} src/manage.py makemigrations rdwatch --check --dry-run
    {envpython} src/manage.py makemigrations rdwatch_scoring --check --dry-run
```

django/src/rdwatch_scoring/router.py
```diff
@@ -0,0 +1,40 @@
...
+        """
+        Attempts to write scoringdb models go to auth_db.
+        """
+        if model._meta.app_label in self.route_app_labels:
+            return 'scoringdb'
+        return None
+
+    def allow_relation(self, obj1, obj2, **hints):
+        """
+        Allow relations if a model in the scorin apps is
```
mvandenburgh, line 27:
```suggestion
        Allow relations if a model in the scoring app is
```

mvandenburgh (changes requested):
Just an initial pass I made from reading through the code - next, I'll test it out and provide more feedback there.

django/src/rdwatch/api.py
```diff
@@ -10,4 +12,6 @@
 api.add_router('/performers/', performer_router)
 api.add_router('/regions/', region_router)
 api.add_router('/model-runs/', site_model_router)
+# Note lets not use this, try to implement it in the rdwatch_scoring app
+api.add_router('/scores/', scores)
```
mvandenburgh, line 16: I believe we can remove this now, since these routes are now defined in the `api.py` and `urls.py` of the `rdwatch_scoring` app.
```suggestion
```

django/src/rdwatch/server/settings.py
```diff
@@ -32,8 +32,34 @@ class BaseConfiguration(Configuration):
...
+        DB_val = values.DatabaseURLValue(
+            alias='default',
+            environ_name='POSTGRESQL_URI',
+            environ_prefix=_environ_prefix,
+            environ_required=True,
+            # Additional kwargs to DatabaseURLValue are passed to dj-database-url
+            engine='django.contrib.gis.db.backends.postgis',
+        )
+        db_dict = DB_val.value
+        if os.environ['RDWATCH_POSTGRESQL_SCORING_URI'] is not None:
```
mvandenburgh, line 50: This will crash with a `KeyError` if `'RDWATCH_POSTGRESQL_SCORING_URI'` is not in os.environ.
```suggestion
        if 'RDWATCH_POSTGRESQL_SCORING_URI' in os.environ:
```

django/src/rdwatch_scoring/views.py
```diff
@@ -0,0 +1,216 @@
...
+class EvaluationResponseSchema(Schema):
+    region_name: str = None
+    evaluationId: int = None
+    evaluation_run: int = None
+    performer: str = None
+    evaluation_uuid: str = None
+    unionArea: float = None
+    statusAnnotated: str = None
+    temporalIOU: TemporalSchema = None
+    color: str = None
```
mvandenburgh, line 91: A type annotation like this
`site_preparation: str = None`
is actually invalid and would fail type-checking by a modern python type checker. Instead, if the variable truly can be `None`, we should do
`site_preparation: str | None = None`.
If the variable is never `None` and will always be a string, then we should remove the assignment and just do
`site_preparation: str`

Not that we're using a type-checker here, but I think it's still worth keeping our types correct.

django/src/rdwatch_scoring/views.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+# If the combination of configurationId and regionId has scoring
+@router.get('/has_scores')
+def has_scores(request: HttpRequest, configurationId: int, regionId: int):
+    # from the hyper parameters we need the evaluation and evaluation_run Ids
+    configuration = get_object_or_404(HyperParameters, pk=configurationId)
+    evaluationId = configuration.evaluation
+    performer_name = configuration.performer.slug.lower()
+    evaluation_run = configuration.evaluation_run
```
mvandenburgh, line 101: The ORM will generate two SQL queries here, when we can do it in one with [`select_related`](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#select-related):
```suggestion
    configuration = get_object_or_404(HyperParameters.objects.select_related('performer'), pk=configurationId)
    evaluationId = configuration.evaluation
    performer_name = configuration.performer.slug.lower()
    evaluation_run = configuration.evaluation_run
```

django/src/rdwatch_scoring/views.py
```diff
@@ -0,0 +1,216 @@
...
+    if evaluation:
+        evaluationUUID = evaluation.uuid
+        # Now we can get a list of the  SiteIds information for this evaluationId
+        site_list = Site.objects.filter(
+            region_id=region_name,
+            evaluation_run_uuid=evaluationUUID,
+            originator=performer_name,
+        )
+        colorMappers = {}
+        for _i, site in enumerate(site_list):
```
mvandenburgh, line 137: It doesn't appear that the counter variable is used here, so I think we can remove this enumerate call
```suggestion
        for site in site_list:
```

django/src/rdwatch_scoring/views.py
```diff
@@ -0,0 +1,216 @@
...
+    performer: str = None
+    evaluation_uuid: str = None
+    unionArea: float = None
+    statusAnnotated: str = None
+    temporalIOU: TemporalSchema = None
+    color: str = None
+
+
+# If the combination of configurationId and regionId has scoring
+@router.get('/has_scores')
```
mvandenburgh, line 95: The convention elsewhere in the app has been to use dashes instead of underscores in urls:
```suggestion
@router.get('/has-scores')
```

django/src/rdwatch_scoring/views.py
```diff
@@ -0,0 +1,216 @@
...
+    return EvaluationRun.objects.filter(
+        evaluation_run_number=evaluation_run,
+        evaluation_number=evaluationId,
+        region=region_name,
+        performer=performer_name,
+    ).exists()
+
+
+# Region scoring Map
+@router.get('/region_scoring_colors')
```
mvandenburgh, line 113: Considering the full endpoint URL will be `/api/scores/region_scoring_colors`, what do you think about removing the reference to scoring here?
```suggestion
@router.get('/region-colors')
```

django/src/rdwatch_scoring/views.py
```diff
@@ -0,0 +1,216 @@
...
+            colorMappers[
+                f'{configurationId}_{regionId}_{configuration.performer.id}_{site_number}'  # noqa: E501
+            ] = get_site_scoring_color(evaluationUUID, 'overall', site.site_id)
+
+    return colorMappers
+
+
+# Gets the basic scoring information based on the
+# configurationId, regionId SiteNumber and Version
+@router.get('/scoring_details', response=EvaluationResponseSchema)
```
mvandenburgh, line 152: Considering the full endpoint URL will be `/api/scores/scoring_details`, what do you think about removing the reference to scoring here?
```suggestion
@router.get('/details')
```
Or perhaps even
```suggestion
@router.get('/')
```

scripts/loadGroundTruth.py
```diff
@@ -0,0 +1,88 @@
...
+                post_model_data = {
+                    "performer": "TE",
+                    "title": "Ground Truth",
+                    "parameters": {},
+                }
+                if expiration_time is not None:
+                    post_model_data["expiration_time"] = expiration_time
+
+                res = requests.post(
+                    f"{rgd_endpoint}/api/model-runs",
```
mvandenburgh, line 65: Updated as of #222 
```suggestion
                    f"{rgd_endpoint}/api/model-runs/",
```

scripts/loadGroundTruth.py
```diff
@@ -0,0 +1,88 @@
...
+                    cookies=cookies,
+                )
+                res.raise_for_status()
+                existing_model_runs[model_run_name] = res.json()["id"]
+
+            model_run_id = existing_model_runs[model_run_name]
+
+            print(f"  Posting {file}...")
+            res = requests.post(
+                f"{rgd_endpoint}/api/model-runs/{model_run_id}/{endpoint}",
```
mvandenburgh, line 77: Updated as of #222 
```suggestion
                f"{rgd_endpoint}/api/model-runs/{model_run_id}/{endpoint}/",
```

scripts/loadModelRun.py
```diff
@@ -0,0 +1,141 @@
...
+    site_models_glob,
+    rgd_auth_cookie,
+    title="Ground Truth",
+    performer_shortcode="TE",
+    eval_num=None,
+    eval_run_num=None,
+    expiration_time=None,
+):
+    # Check that our run doesn't already exist
+    model_run_results_url = f"{rgd_endpoint}/api/model-runs"
```
mvandenburgh, line 60: Updated as of #222 
```suggestion
    model_run_results_url = f"{rgd_endpoint}/api/model-runs/"
```

scripts/loadModelRun.py
```diff
@@ -0,0 +1,141 @@
...
+            and model_run["region"] is not None
+            and model_run["region"].get("name") == region_id
+        ):  # noqa
+            existing_model_run = model_run
+            break
+
+    if existing_model_run is not None:
+        model_run_id = model_run["id"]
+    else:
+        post_model_url = f"{rgd_endpoint}/api/model-runs"
```
mvandenburgh, line 86: Updated as of #222 
```suggestion
        post_model_url = f"{rgd_endpoint}/api/model-runs/"
```

scripts/loadModelRun.py
```diff
@@ -0,0 +1,141 @@
...
+
+        post_model_result = requests.post(
+            post_model_url,
+            json=post_model_data,
+            headers={"Content-Type": "application/json"},
+            cookies=cookies,
+        )
+        model_run_id = post_model_result.json()["id"]
+
+    post_site_url = f"{rgd_endpoint}/api/model-runs/{model_run_id}/site-model"
```
mvandenburgh, line 110: Updated as of #222 
```suggestion
    post_site_url = f"{rgd_endpoint}/api/model-runs/{model_run_id}/site-model/"
```

mvandenburgh (changes requested):

docker-compose.override.yaml
```diff
@@ -65,6 +65,17 @@ services:
...
+  scoredb:
+    image: postgis/postgis:latest
+    environment:
+      POSTGRES_USER: scoring
+      POSTGRES_DB: scoring
+      POSTGRES_PASSWORD: secretkey
+    volumes:
+      - postgresql-scoring-data:/var/lib/postgresql/data
+    ports:
+      - 5433:5432
```
mvandenburgh, line 79: Can this be moved to `docker-compose.yaml`? In the current setup, if one is running just `docker-compose.yaml` for postgres/rabbit/minio etc, and running django/celery natively, then there is no way to run the scoring DB.

mvandenburgh (approved):
I left one minor change request for the docker-compose set up, but once that is addressed this looks good to me!

DEVELOPMENT.md
```diff
@@ -94,3 +94,56 @@ Services the application requires.
...
+In development mode a scoring Database is automatically initialized at URI: `postgresql+psycopg2://scoring:secretkey@localhost:5433/scoring`
+
+To score data:
+- Begin by first starting the the score DB using `docker compose up scoredb`
+- After that in the Metrics and Test Framework, copy the `alembic_example.ini` to `alembic.ini` and set the `sqlalchemy.url = postgresql+psycopg2://scoring:secretkey@localhost:5433/scoring`
+- install the scoring code `pip install -e .` from the root of the Metrics and Test Framework Repository
+- run `alembic upgrade head` to make sure the database is configured properly to load results.
+    - Instead of running the above command you can run the following command from inside the RD-WATCH repository `docker compose run --rm django poetry run django-admin migrate rdwatch_scoring --database scoringdb` this will use the django migrations to initialize the database.
+- Execute the scoring code from inside the mtrics and test framework:
+```
```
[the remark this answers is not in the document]
BryonLewis: I mean really all that AWS needs is a DB with a specific schema.  Having the repo may make it easier to run the initialization of the DB in AWS.

If we have the entire scoring we could also make it so on upload we could conditionally score, or run scoring after the fact on a model_run.  (The repo includes the ground truth).

dev/.env.docker-compose-native
```diff
@@ -10,3 +10,4 @@ AWS_REQUEST_PAYER=requester
 RDWATCH_POSTGRESQL_URI="postgresql://rdwatch:secretkey@localhost:5432/rdwatch"
 RDWATCH_REDIS_URI="redis://:secretkey@localhost:6379"
 RDWATCH_CELERY_BROKER_URL="amqp://rdwatch:secretkey@localhost:5672/"
+RDWATCH_POSTGRESQL_SCORING_URI="postgresql://scoring:secretkey@scoredb:5432/scoring"
```
mvandenburgh, line 13: One more issue i just noticed when testing this with native django
```suggestion
RDWATCH_POSTGRESQL_SCORING_URI="postgresql://scoring:secretkey@localhost:5433/scoring"
```

django/tox.ini
```diff
@@ -18,6 +18,7 @@ passenv =
     RDWATCH_DJANGO_DEBUG
     RDWATCH_SMART_STAC_URL
     RDWATCH_SMART_STAC_KEY
+    RDWATCH_POSTGRESQL_SCORING_URI
```
mvandenburgh, line 21: Also, I don't think we need to include this here anymore now that the `DATABASES` setting is a property and can handle `RDWATCH_POSTGRESQL_SCORING_URI` not being set. In fact, doing so causes tests to pass again for me, at least locally.
```suggestion
```

docker-compose.override.yaml
BryonLewis: yeah, I moved the other container into the main docker-compose.

django/tox.ini
BryonLewis: Thanks for finding that.  Yeah, that must make only the default DB load so the tests no longer have an issue determining which DB to run on.  I looked up and tried some settings to set them manually but it didn't seem to work properly.

PR #216, opened 2023-06-12
Add celery beat

mvandenburgh:
This PR 
- Moves all non-task functions out of `rdwatch.tasks` and into their own module
	- This is the usual practice when it comes to celery tasks AFAIK, and it also makes it easier to see all celery tasks at a glance. @BryonLewis any objections to this?
- Adds a new docker compose service for celery beat and instructions for running it natively.
- Adds a stub task that runs once a minute. I will make a follow up to actually implement the "temporary" model run logic, where this task will be responsible for querying for and deleting model runs that need to be deleted.

### DIFF
--- a/django/src/rdwatch/server/settings.py
+++ b/django/src/rdwatch/server/settings.py
@@ -1,4 +1,5 @@
 import os
+from datetime import timedelta
 
 from configurations import Configuration, values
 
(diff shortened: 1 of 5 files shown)


django/src/rdwatch/server/settings.py
```diff
@@ -93,6 +94,13 @@ class BaseConfiguration(Configuration):
     CELERY_RESULT_EXTENDED = True
     CELERYD_TIME_LIMIT = 1200
 
+    CELERY_BEAT_SCHEDULE = {
+        'delete-temp-model-runs-beat': {
+            'task': 'rdwatch.tasks.delete_temp_model_runs_task',
+            'schedule': timedelta(seconds=60),
```
BryonLewis, line 100: for a dummy task do we want to bump this up to like once an hour or day (outside of testing?).  Maybe a configurable time that changes based on deployment/development?

BryonLewis (approved):
I didn't pull and test directly, but I don't see anything that would worry me.

django/src/rdwatch/server/settings.py
mvandenburgh: I changed it to an hour for now. But yeah we need to decide on the exact semantics of the "expiration time" of a model run, and adjust the interval accordingly.

PR #217, opened 2023-06-13
Fixes Tooltips and NavBars

BryonLewis:
The Tooltips weren't working before in the ./dist version of the code because of some mounting issues that weren't the same when running a local vite server.

To duplicate the issue first:

- build the web web app `cd ./vue` && `yarn build`
- install something like `yarn global add http-server`, then run `http-server -P http://localhost:8000 ./dist` to start the server
- With the main branch you'll notice the system doesn't display the tooltip.
- Try the same process again with the updated code and you should see the tooltip.

I also did the DockerFile and nginx.json modifications so I could build this version.  Just confirm this will work properly.

Finally, I added a `permanent` directive to the `v-navigation-bar`s to prevent collapsing.  In the future, we can look into having a small and wide version of the navigation bars.

### DIFF
--- a/vue/src/interactions/mouseEvents.ts
+++ b/vue/src/interactions/mouseEvents.ts
@@ -1,28 +1,11 @@
-import { App, Ref, createApp, defineComponent, nextTick, ref } from "vue";
+import { App, Ref,   nextTick, ref } from "vue";
 import { Color, Map, MapLayerMouseEvent, Popup } from "maplibre-gl";
 import { ShallowRef } from "vue";
 import { getSiteObservationDetails, selectedObservationList, state } from "../store";
-import PopUpComponent from '../components/PopUpComponent.vue'
+import  createPopup from '../main';
+import { PopUpData } from '../interactions/popUpType';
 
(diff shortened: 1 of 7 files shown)


vue/src/interactions/mouseEvents.ts
```diff
@@ -96,21 +79,14 @@ const popupLogic = (map: ShallowRef<null | Map>) => {
...
       popup.setLngLat(coordinates).setHTML('<div id="popup-content"></div>').addTo(map.value);
-      const MyNewPopup = defineComponent({
-        extends: PopUpComponent,
-        setup() {
-          return { data: popupData }
-        },
-      })
       nextTick(() => {
-        if (app !== null) {
+        if (app != null) {
```
mvandenburgh, line 83: Why change to a less strict equality check?

BryonLewis: that was a mistake, I'll fix it and I'm surprised the linter didn't yell at me.

mvandenburgh (approved):

PR #222, opened 2023-06-16
Convert model run endpoints to ninja

mvandenburgh:


django/src/rdwatch/views/model_run.py
```diff
@@ -158,96 +163,136 @@ def get_queryset():
...
+def post_region_model(
+    request: HttpRequest,
+    hyper_parameters_id: int,
+    region_model: RegionModel,
+):
+    hyper_parameters = get_object_or_404(HyperParameters, pk=hyper_parameters_id)
+    site_evaluations = SiteEvaluation.bulk_create_from_from_region_model(
+        region_model, hyper_parameters
+    )
+    return 201, [eval.id for eval in site_evaluations]
```
mvandenburgh, line 298: As part of this, I moved these two endpoints from `site_model.py` since they fall under the same model run router.

BryonLewis (changes requested):
The route for posting model-runs doesn't seem to be working.

PR #223, opened 2023-06-20
Auto-deletion of model runs

mvandenburgh:
This PR adds `created` and `expiration_time` fields to the HyperParameters model. The `expiration_time` is set by the user at model run creation time, and is measured in hours. An hourly celery beat task checks for any "expired" model runs and deletes them + all of their evaluations and observations.

### DIFF
--- a/django/src/rdwatch/models/hyper_parameters.py
+++ b/django/src/rdwatch/models/hyper_parameters.py
@@ -1,7 +1,10 @@
+from django_extensions.db.models import CreationDateTimeField
+
 from django.db import models
 
 
 class HyperParameters(models.Model):
+    created = CreationDateTimeField()
     title = models.CharField(max_length=1000)
     performer = models.ForeignKey(
         to='Performer',
@@ -12,6 +15,11 @@ class HyperParameters(models.Model):
     parameters = models.JSONField(
         help_text='The hyper parameters for an ML task',
         db_index=True,
+    )
+    expiration_time = models.DurationField(
+        null=True,
+        blank=True,
+        help_text='Time relative to creation that this model run should be deleted.',
     )
 
     def __str__(self):
--- a/django/src/rdwatch/serializers/hyper_parameters.py
+++ b/django/src/rdwatch/serializers/hyper_parameters.py
@@ -10,6 +10,9 @@ class HyperParametersWriteSerializer(serializers.Serializer):
     performer = serializers.CharField()
     title = serializers.CharField(max_length=1000)
     parameters = serializers.JSONField(default=dict)
+    expiration_time = serializers.IntegerField(
+        min_value=0, max_value=24 * 365, required=False
+    )
 
     def validate_performer(self, value: str) -> lookups.Performer:
         try:
--- a/django/src/rdwatch/serializers/hyper_parameters.py
+++ b/django/src/rdwatch/serializers/hyper_parameters.py
@@ -39,6 +39,8 @@ class HyperParametersDetailSerializer(serializers.Serializer):
     timestamp = serializers.IntegerField(allow_null=True)
     timerange = TimeRangeSerializer(allow_null=True)
     bbox = serializers.JSONField(allow_null=True)
+    created = serializers.DateTimeField()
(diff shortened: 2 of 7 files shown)


django/src/rdwatch/models/hyper_parameters.py
```diff
@@ -13,6 +16,11 @@ class HyperParameters(models.Model):
         help_text='The hyper parameters for an ML task',
         db_index=True,
     )
+    expiration_time = models.DurationField(
+        null=True,
+        blank=True,
+        help_text='Time relative to creation that this model run should be deleted.',
+    )
```
BryonLewis, line 23: While this is null/blank.  I think the POST endpoint requires that there is an `expiration_time`.  Might want to sync these up, or come up with some reasonable default time?

django/src/rdwatch/views/model_run.py
```diff
@@ -165,6 +167,9 @@ def create(self, request: Request):
             title=write_serializer.validated_data['title'],
             performer=write_serializer.validated_data['performer'],
             parameters=write_serializer.validated_data['parameters'],
+            expiration_time=timedelta(
+                hours=write_serializer.validated_data['expiration_time']
+            ),
```
BryonLewis, line 172: In this same file, it may be a good idea to include the expiration_time in the /model-runs endpoint so the client can display when the model-run expires/deletes.

django/src/rdwatch/serializers/hyper_parameters.py
```diff
@@ -10,6 +10,9 @@ class HyperParametersWriteSerializer(serializers.Serializer):
     performer = serializers.CharField()
     title = serializers.CharField(max_length=1000)
     parameters = serializers.JSONField(default=dict)
+    expiration_time = serializers.IntegerField(
+        min_value=0, max_value=24 * 365, required=False
```
BryonLewis, line 14: While I see that this has required=false, if you don't include it you get an error when posting models.

django/src/rdwatch/tasks/__init__.py
```diff
@@ -206,5 +215,15 @@ def get_siteobservations_images(
 @shared_task
 def delete_temp_model_runs_task() -> None:
     """Delete all model runs that are due to be deleted."""
-    print('Test')
-    # TODO: implement this
+    model_runs_to_delete = HyperParameters.objects.alias(
+        delete_at=ExpressionWrapper(
+            F('created') + F('expiration_time'), output_field=DateTimeField()
+        )
+    ).filter(delete_at__lte=timezone.now())
```
BryonLewis, line 222: What is the behavior of this if the expiration_time is null/blank?  Should there be a specific value that prevents deletion (-1 or 0 or something else)?

BryonLewis (changes requested):
I know there seems to be some stuff in there so you don't need to require an expiration_time during post, but that doesn't seem to work in practice.

I think we need the determination if there should be a default non-delete value and to skip or determine the action if there is a null/default value when the beat task is run.

django/src/rdwatch/models/hyper_parameters.py
mvandenburgh: Yeah that was an oversight, we do want to allow null/blank here because not all model runs should have an expiration date (Ground truths for example, we never want deleted)

django/src/rdwatch/tasks/__init__.py
BryonLewis: Yeah, I didn't actually configure a null row in the table to verify, but just wanted to double-check.  I know that with a time of 0 and setting the update rate to 1 minute it would immediately delete it.  The explicit thing should make it easier to reason about what's going on when someone else looks at it in the future, thanks.

django/src/rdwatch/views/model_run.py
mvandenburgh: :+1: https://github.com/ResonantGeoData/RD-WATCH/pull/223/commits/568a891d822ff97bf2d02c9945d6cdd8b6252374

BryonLewis (approved):
:+1:

PR #225, opened 2023-06-21
Convert vector tile endpoint to ninja

mvandenburgh:
Doing this first before integrating caching, since presumably we'll want to use ninja's native caching mechanism.

### DIFF
--- a/django/src/rdwatch/api.py
+++ b/django/src/rdwatch/api.py
@@ -3,9 +3,11 @@ from ninja import NinjaAPI
 from .views.performer import router as performer_router
 from .views.region import router as region_router
 from .views.site_model import router as site_model_router
+from .views.vector_tile import router as vector_tile_router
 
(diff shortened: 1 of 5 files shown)


BryonLewis (approved):
Pulled and tested and this looks good.

Just making some notes for the future:

- Maybe we want to break out the region into its own endpoint separate from the vector tiles.  That way we don't have to do an intersection calculation on all regions if they aren't necessary.
- The client side shouldn't probably add the observation/evaluation source until one of the them is selected.  This prevents it from trying to grab MB of data with nothing visible.
- The client site might also benefit from only turning on the vector tile after it is zoomed into some specific range to prevent multiple cascading vector calculations at the zooming moves in.
- Adding an initial pre-filter with an array of configurationIds I think should make the endpoint much more responsive than it is by quickly filtering out items that don't need to be displayed instead of doing an intersection calculation on them.

PR #226, opened 2023-06-26
Limit vector tiles by select model runs

mvandenburgh:
Backend changes:
- Move the vector tiles endpoint underneath the `model-runs` prefix (i.e. the endpoint is now `/api/model-runs/<id>/vector-tile/z/x/y.pbf`) and require a model run ID to filter by. The vector tiles are then generated based on the evaluation/observation/region polygons associated with that model run, as opposed to every polygon in the database that intersects with the given coordinates.

Frontend changes:
- Move model runs / opened model runs into the global state
- Refactor the `rdwatchtiles` to create a new vector tiles layer for every selected model run. This means that vector tiles will only be fetched for the currently selected model runs.

### DIFF
--- a/django/src/rdwatch/views/vector_tile.py
+++ b/django/src/rdwatch/views/vector_tile.py
@@ -1,5 +1,3 @@
-from ninja import Router
-
 from django.contrib.gis.db.models.functions import Area, Transform
 from django.db import connection
 from django.db.models import (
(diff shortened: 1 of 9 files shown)


mvandenburgh:
Upon further testing, it looks like this breaks the mouseover popup on the individual sites - I'll mark this back to draft.

mvandenburgh:
@BryonLewis this ready for review. I addressed the refetching issue discussed this morning in https://github.com/ResonantGeoData/RD-WATCH/pull/226/commits/cef1f346569979b4721df175fdc090cd63ceb441. I'll make a follow on PR adding caching to this, as theres still a little bit of room for improvement in terms of response times imo

django/src/rdwatch/views/vector_tile.py
```diff
@@ -42,7 +20,7 @@ def vector_tile(request: HttpRequest, z: int, x: int, y: int):
     )
 
     evaluations_queryset = (
-        SiteEvaluation.objects.filter(intersects)
```
BryonLewis, line 45: Should we keep the filter for intersection as a secondary filter?
 I.E SiteEvaluation.objects.filter(configuration_id=hyper_parameters_id, intersects)`
We don't have many but if a model-run spans a large area it will cull some of polygons returned outside of the view.  I believe as long as the order is maintained like above it would do the configuration_id first and then the intersections

BryonLewis (approved):
This is great, I did some testing to see how much it reduces the data and number of requests and it is significantly less than before.

I have one comment about if we should leave the intersects calculation in there as another filter.  It doesn't need to be added back in but it may be a minor improvement when zooming in far on polygons.

django/src/rdwatch/views/vector_tile.py
mvandenburgh: Good point, I added this back in.

PR #230, opened 2023-07-11
Add temporary backwards-compatible endpoints

mvandenburgh:
This is a temporary feature that will help consumers of the API to migrate to the new trailing slash endpoints.


mvandenburgh:
@BryonLewis I took another shot at this and got it to work - I propose that we merge this and do a deploy, then let everyone know about the breaking change, and then we can remove them in the next deployment.

BryonLewis (approved):
:+1:

PR #233, opened 2023-07-18
Add timestamp-based caching for vector tiles endpoint

mvandenburgh:
This PR adds caching for the vector tiles endpoint. Previously, we had to remove the caching because it was difficult to keep the cache up to date, since uploading new site evaluations/observations to a model run would invalidate the current cache. This PR solves this by caching the tiles based on 
a) The model run ID 
b) The vector tile coordinates
c) The latest _created_ timestamp between the model run's creation date, each associated site evaluation's creation date, and each associated site observation's creation date.
	- To enable this, this PR also adds `created` fields to the `SiteEvalutation` and `SiteObservation` models.


django/src/rdwatch/models/site_evaluation.py
```diff
@@ -15,6 +16,7 @@
 
 
 class SiteEvaluation(models.Model):
+    created = CreationDateTimeField()
```
BryonLewis, line 19: Is the timestamp field technically a created time already?
https://github.com/ResonantGeoData/RD-WATCH/blob/main/django/src/rdwatch/models/site_evaluation.py#L76

django/src/rdwatch/models/site_observation.py
```diff
@@ -11,6 +12,7 @@
 
 
 class SiteObservation(models.Model):
+    created = CreationDateTimeField()
```
BryonLewis, line 15: In my other comment about the timestamp for the `siteEvaluation` being like a 'created' date.  Each `SiteObservation` has a `siteeval_id` which would link it to a `siteEvaluation` and that has a timestamp that is created when it is uploaded.

BryonLewis (changes requested):
Do we need siteEval and SiteObs timestamps as an additional check because there is a use case where users add more annotations to a model-run after initial upload?

Looking through it, I was wondering if we have enough information already without adding a CreationDateTimeField.  It looks like the siteEvaluation timestamp is already a updated time that we use.
image
https://github.com/ResonantGeoData/RD-WATCH/blob/main/django/src/rdwatch/views/model_run.py#L160

This datetime is retrieved by getting the max timestamp from the evaluations already.  Each SiteObservation has a siteeval_id which references a SiteEvaluation which would have the timestamp as well.

Just have questions about it, I could very well see the need to have created fields for both siteEvaluations and SiteObservations and how they could be useful for future filtering.

The caching may be able to be simplified if we just use the timestamp as an additional query parameter since the client has it already in the model run when they enable it.  I.E instead of just sending the hyperparmeter_id with the x/y/z we could send the timestamp as well.  As more siteObservations are added to the model run, the client date/time would update automatically.

Like I said these are just thoughts and I'm open to doing it either way just wanted to propose this method.

django/src/rdwatch/models/site_evaluation.py
mvandenburgh: Good catch. I changed it to use that instead and backed out the `created` change.

mvandenburgh:
> The caching may be able to be simplified if we just use the timestamp as an additional query parameter since the client has it already in the model run when they enable it. I.E instead of just sending the hyperparmeter_id with the x/y/z we could send the timestamp as well. As more siteObservations are added to the model run, the client date/time would update automatically.

My thinking on this is that it's better for it to be opaque to the server. Thinking in terms of separation-of-concerns between the client and server, I think it should be the server's responsibility to provide up-to-date vector tiles. Not that we're trying to support any clients besides the web frontend, but I think from an API design perspective, the idea that you could upload a site model and then get back incorrect vector tiles (which is really just invalid/corrupted data at that point) due to specifying an incorrect time stamp seems wrong.

mvandenburgh:
The tests I added are failing because Redis isn't hooked up properly in CI; I did that and also went ahead and removed the multi-node cluster stuff in #237, as we're not using that anymore in production.

mvandenburgh:
CI is passing now that #237 has been merged into this branch.

BryonLewis (approved):
:+1:

PR #234, opened 2023-07-18
Convert server status endpoint to ninja

mvandenburgh:


django/src/rdwatch/views/server_status.py
```diff
@@ -1,70 +1,35 @@
...
+    days: int
+    seconds: int
+    useconds: int = Field(..., alias='microseconds')
 
-class ServerStatusSchema(AutoSchema):
-    def get_operation_id(self, *args):
-        return 'getStatus'
+    @staticmethod
+    def resolve_iso8601(obj: datetime.timedelta):
+        return str(obj)
```
mvandenburgh, line 21: Note, this returns a slightly different string than the DRF serializer for `datetime.timedelta`. Previously, the return value would look something like `6 05:04:40.740791`, but now it is `6 days, 05:04:40.740791`.

mvandenburgh:
@BryonLewis with #243 being merged, this is now the last Ninja conversion. I'll make a follow up PR to this that removes the remaining bits of DRF (mostly DRF exceptions still being used).

django/src/rdwatch/views/server_status.py
BryonLewis: In testing, I'm getting back an iso8601 duration format:  `"iso8601": "P0DT00H03M56.226221S"` which I think is fine.

BryonLewis (approved):
Looks fine to me.  I may update this endpoint in the future to include capabilities (Like knowing that there is a scoring database connected).

PR #235, opened 2023-07-20
Enable Django admin

mvandenburgh:
Adds the needed settings in order to enable the Django admin at `/admin`. I sometimes find this easier to use then `shell_plus` or looking at the database itself.

Settings needed to enable this were taken from
https://github.com/girder/django-composed-configuration

### DIFF
--- a/django/src/rdwatch/server/settings.py
+++ b/django/src/rdwatch/server/settings.py
@@ -3,7 +3,7 @@ from datetime import timedelta
 
 from configurations import Configuration, values
 
-_environ_prefix = 'RDWATCH'
+_ENVIRON_PREFIX = 'RDWATCH'
 
 
 # With the default "late_binding=False", and "environ_name" is
@@ -20,7 +20,7 @@ class BaseConfiguration(Configuration):
     USE_X_FORWARDED_HOST = True
     WSGI_APPLICATION = 'rdwatch.server.application'
     ALLOWED_HOSTS = ['*']
-    DEBUG = values.BooleanValue(False, _environ_prefix='RDWATCH_DJANGO')
+    DEBUG = values.BooleanValue(False, environ_prefix='RDWATCH_DJANGO')
     # Django's docs suggest that STATIC_URL should be a relative path,
     # for convenience serving a site on a subpath.
     STATIC_URL = 'static/'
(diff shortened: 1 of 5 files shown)


mvandenburgh:
Note this will not expose the admin on the AWS deployment, as NGINX only proxies backend requests to `/api`, and not `/admin`.

BryonLewis:
Am I missing something in this to get actual template rendering so the admin interface isn't just raw ugly HTML without styling?

django/src/rdwatch/server/settings.py
```diff
@@ -21,12 +21,19 @@ class BaseConfiguration(Configuration):
     WSGI_APPLICATION = 'rdwatch.server.application'
     ALLOWED_HOSTS = ['*']
     DEBUG = values.BooleanValue(False, _environ_prefix='RDWATCH_DJANGO')
```
BryonLewis, line 23: Not sure what is going on, but this isn't evaluating to True when RDWATCH_DJANGO_DEBUG is set to 1 on my system.  Looking at all os.environ I can see it in there, but this is always returning false;
```suggestion
    DEBUG = 'RDWATCH_DJANGO_DEBUG' in os.environ
```

BryonLewis (changes requested):
I like the idea of having this in there to look at stuff quickly instead of utilizing pgadmin or the database directly.

One additional suggestion:
Adding to the docs some instructions about createsuperuser for the admin user docker: `docker compose run --rm django poetry run django-admin createsuperuser`  and maybe the regular command for non-docker dev.

django/src/rdwatch/server/settings.py
mvandenburgh: My bad, I fixed this locally but didn't push it here - I opened a PR on top of this branch with the fix #236 as its unrelated to this PR.

mvandenburgh:
Added instructions for `createsuperuser` to the dev docs.

django/src/rdwatch/server/settings.py
BryonLewis: Marking as resolved, your other PR clears it up.

BryonLewis (approved):

PR #236, opened 2023-07-20
Fix typo

mvandenburgh:
This typo resulted in the `DEBUG` setting not being read from the `RDWATCH_DJANGO_DEBUG` env var as expected, meaning that all dev environments are likely (incorrectly) being configured with `DEBUG = False`

This also causes the CSS issues in #


BryonLewis (approved):
I was trying different values ('y', 'yes', 1, '1') and stuff for like 10 minutes to figure out why it wasn't working and didn't notice that extra underscore there at all.

PR #237, opened 2023-07-20
Switch to native django redis adapter + docker image

mvandenburgh:
Now that we're no longer using multi-node Redis (see https://smartgitlab.com/kitware/rgd_aws_infrastructure/-/commit/d4b077e6afc5376c23d310e182293e8ea0e85d18), we can switch back to the vanilla Django redis adapter and remove the hack we had to get the `django-redis` package to work. I've also updated the local dev environment to use the official redis docker image instead of the bitnami cluster one.


docker-compose.override.yaml
```diff
@@ -50,7 +50,7 @@ services:
       - ./dev/.env.docker-compose
     environment:
       RDWATCH_POSTGRESQL_URI: "postgresql://rdwatch:secretkey@postgresql:5432/rdwatch"
-      RDWATCH_REDIS_URI: "redis://:secretkey@redis-master:6379"
+      RDWATCH_REDIS_URI: "redis://:redis:6379"
```
BryonLewis, line 53: Celery Worker wouldn't connect in docker dev mode
```suggestion
      RDWATCH_REDIS_URI: "redis://redis:6379"
```

BryonLewis (changes requested):
One minor change to the docker-compose.override.yaml

BryonLewis (approved):

PR #238, opened 2023-07-21
Remove unused view

mvandenburgh:
This view was replaced at some point by the views in `site_evaluation.py` and `site_observation.py` and isn't used anywhere, or even hooked up in `urls.py`.


BryonLewis (approved):
:+1:

PR #239, opened 2023-07-21
Convert site evaluations endpoint to Ninja

mvandenburgh:


mvandenburgh:
It doesn't look like this endpoint is used at all by the client, so in theory we could remove it... although I think it's still good to expose this data in an endpoint. @BryonLewis any thoughts?

BryonLewis:
I'd say leave it because we may need it for the modification of siteEvaluations.

BryonLewis (approved):

PR #240, opened 2023-07-24
Fix failing deploy due to `CACHES` change

mvandenburgh:
The `CacheURLValue` class from `django-configurations` doesn't allow specifying a key_prefix, which was causing auth errors between the Django app server and Redis for some reason. This PR updates that setting to explicitly specify the `CACHES` dict instead. It also removes the `cache` extra from the `django-configurations` dependency as we no longer need it now.


mvandenburgh:
@BryonLewis pinging so you're aware. I'll merge this and deploy after tests pass.

PR #244, opened 2023-07-25
Site Observation Editor

BryonLewis:
Demo:
https://github.com/ResonantGeoData/RD-WATCH/assets/61746913/b8a8fec2-b449-48a1-9c7a-2db4162444f1

Tasks:
- [x]  Create sample data with null timestamps
- [x] Modify models to support null timestamps
- [x] Modify Image downloading so it will work with null timestamps and return a min/max of images between 2013-01-01 and now.
- [x] Update client-side rendering of dates/times and vector tiles to support null timestamps.
- [x] Add SiteEvaluation Label to image display endpoint
- [x] Refactor siteObservation view to utilize ninja
- [x] Add Notes category to SiteObservations/SiteEvaluations
- [x] Give a startDate/endDate to the siteEvaluations that can be set to null if needed.  This prevents the dates for a siteevaluation being dependent on the times for siteObservation.
- [x] Create New Table to hold modifications to siteObservations.  Stuff like a timestamp and a copy of the site observation model along with an indicator of the change (modified, deleted, added).  Have this indexed by siteEvaluation so there is a reference to it.
- [x] When ingesting models it should be indicated that these models are proposed site edits and that when creating them it should create a list of proposed Sites in  table which can be indexed.  It should automatically mark a model run as proposal if it contains regions with 9999.
- [x] SiteEvaluations will have a status that indicates information about it.  Null is default, PROPOSAL means it is a proposed site, APPROVED means it is adjudicated and approved, REJECTED means that it is rejects and not approved.  This is used as  a quick index to view the state of SiteEvaluations.
- [x] Allow users to select a custom date range when getting images so they can override the default values of the site range
- [x] Ground truth data needs to be displayed in parallel with proposed site observation.
- [x] The /Annotation route on the client should only show model-runs which are proposed sites and change /annotation to /proposals
- [x] Make it so you can have siteModels (SiteEvaluation) with no observations.  It needs to use the polygon from the SiteEvaluation and the label should just empty without display/colorl.
- [x] When a site observation is no-label (Or check for null values) make sure it shows no-activity
- [x] Add POST export functionality which will export the model-run in the geoJSON format

Changes:

Database Changes:

- HyperParameters: add ground_truth utilizing the validated field, also added proposal to know that a modelRun is a set of proposals.
- **SiteEvaluation** - Add new cache values (cache_originator, cache_hash, cache_timestamp).  These are for exporting and maintaining this data in the JSON.
- **SiteEvaluation** - Add start/end dates so that siteEvaluations have information about their start/end date now and can support null dates for their start/end date
- **SiteEvaluation** - Add notes field so notes can be added to siteEvaluations
- **SiteEvaluation** - Add 'status' field to indicate the status of SiteEvaluation:  PROPOSAL, APPROVED, REJECTED
- **SiteObservation** - Updated timestamp to allow Null values
- **SiteObservationTracking** - New Model which duplicates a lot of the data for SiteObservation for tracking the changes.  I initially tried an abstract base class to use for SiteObservation and SiteObservationTracking but it works by using references which wouldn't work for our needs.
- **SiteEvaluationTracking** - New Model which duplicates a lot of data for SiteEvaluation for tracking the changes.  

REST EndPoint Changes:

- **/model-runs** - Added a subquery to perform calculations of the counts of proposals remaining in the model-run to be adjudicated.  It defaults to not doing this subquery if the model-run does not have the 'proposal' flag.
- **/model-runs** - the 'timerange' now utilizes the new SiteEvaluation start/end dates.
- **/model-runs** - The cancellation of tasks using revoke() now utilizes revoke(terminate=True) to make sure the process is cancelled.
- **/model-runs** - new endpoint (/{id}/download) which will get all siteEvaluations in JSON form and place them into a zip file for downoading.
- /**model-runs** - the filter now has the ability to filter by proposal status (True/False|Null) as well.
- **/evaluations** - Updated references to timerange to utilize start/end dates.
- **/evaluations** - New endpoint utilizing PATCH to update an evaluation.  While updating it will create a copy of the old value in SiteEvaluationTracking so there is a history of the changes.
- **/evaluations** - new endpoint (/{id}download) which will export the data for the SiteEvaluation in the JSON format.  There is a function that is used for this and for the model-runs to generate the files.
- **/evaluations/images** - added label/status/nodes to the return value.  Also added connection to groundTruth if it exists to the response.  Ground truth checks to see if the proposal ID (KR_0001_0023) matches a siteEvaluation that has groundTruth and displays that data.
- **/observations** - Added support for not having any SiteObservations in a main list, so it returns back the SiteEvaluation bounding box
- **/observations** - Updated the generate images (downloading satellite images) so it can accept more parameters (override the date range, set the timerange skipping for S2/L8, and some noData limits)
- **/observations** - The cancel script for the task now includes `revoke(terminate=True)` to truly cancel the process
- /**observations** - Added PATCH and PUT templates for editing SiteObservations but they aren't connected to anything on the front-end right now

Model/Data Ingestion Changes:

- Support for SiteObservations and SiteEvaluations to now support null timestamps and null start/end dates.
- Will automatically turn proposal flag on for a model run if it contains a siteId of 9999
- Added the proposal flag to the script for loading model-runs so they can be set
- Modified support for SiteIds to be larger than smallInts, that's because some kitware ones are 999901
- Added the cache data from the new specification for SiteObservations (cache_timestamp, cache_originator, cache_hash)

Misc Changes:
- fix CI issue with trying to download old packages
- Changed the timerange schema and the Timerange function to support null values for max/min or the value.

FrontEnd Changes:

- Added new endpoints to APIService.ts
- Created new ImagesDownloadingDialog.vue which can be used to configure the options for downloading.  This is utilized at the modelrun level for downloading images as well as the siteEvaluation level.
- Swapped the URL router from /annotations to /proposals because that seemed like it made more sense
- Added all the support for null timeranges and timestamps
- Defaulted the mapLegend to on and added the references for the siteEvaluation phase labels.
- Code was there but wasn't active for the `updateImageMapSources` which makes it so changes in satellite images is smoother without the blanking
- Fixed the Popup logic which was causing the functions to get added multiple times.
- The ImageViewer has been updated to edit the proper fields and have the capability to show groundTruth if it is included.
- Edited the rdwatchTiles so if no SiteObservation exists it displays the SiteEvaluation geometry
- Moved the ability to toggle satellite images outside of the existing component and into store.ts so that proposal could also toggle site images to get a clearer view of things.


django/src/rdwatch/views/site_observation.py
```diff
@@ -187,14 +246,72 @@ def cancel_site_observation_images(request: HttpRequest, pk: int):
...
+    if data.spectrum:
+        site_observation.spectrum = data.spectrum
+    if data.constellation:
+        site_observation.constellation = lookups.Constellation.objects.get(
+            slug=data.label
+        )
+
+    site_observation.save()
+
+    return site_observation
```
mvandenburgh, line 293: Similar situation here as with site eval above -
```suggestion
    with transaction.atomic():
        site_observation = get_object_or_404(SiteObservation.objects.select_for_update(), pk=id)
        # create a copy of it in the history log
        SiteObservationTracking.objects.create(
            timestamp=site_observation.timestamp,
            geom=site_observation.geom,
            score=site_observation.score,
            siteval=site_observation.siteeval,
            notes=site_observation.notes,
            edited=datetime.now(),
            observation=site_observation.pk,
        )

        if data.label:
            site_observation.label = lookups.ObservationLabel.objects.get(slug=data.label)
        if data.notes:
            site_observation.notes = data.notes
        site_observation.timestamp = data.timestamp
        if data.spectrum:
            site_observation.spectrum = data.spectrum
        if data.constellation:
            site_observation.constellation = lookups.Constellation.objects.get(
                slug=data.label
            )

        site_observation.save()

        return site_observation
```

mvandenburgh (changes requested):
I made an initial pass on the code. Next, I'll clone this branch and test it locally and leave any further feedback I get from that.

django/src/rdwatch/views/model_run.py
```diff
@@ -456,3 +518,29 @@ def get_modelrun_evaluations(request: HttpRequest, hyper_parameters_id: int):
...
+        with zipfile.ZipFile(zip_file_path, 'w') as zipf:
+            for item in file_list:
+                zipf.write(item['filename'], f"{item['siteId']}.json")
+
+        with open(zip_file_path, 'rb') as file:
+            response = HttpResponse(file.read(), content_type='application/zip')
+            response[
+                'Content-Disposition'
+            ] = f'attachment; filename="{model_run_name}.zip"'
+            return response
```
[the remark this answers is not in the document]
BryonLewis: For AE Ground Truth it's like 30 seconds (The largest region/model-run.  For most smaller model runs it's 2-5 seconds like shown in the demo video.  I wondered about moving this into a task where when complete it provides a download.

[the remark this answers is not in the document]
mvandenburgh: Yeah I was going to suggest a task too. Someone could easily DDoS the server by attempting to download several model runs at once.

django/src/rdwatch/models/hyper_parameters.py
```diff
@@ -24,5 +24,20 @@ class HyperParameters(models.Model):
...
+        APPROVED = 'APPROVED'  # proposal is approved and merged into ground truth
+
+    proposal = models.CharField(
+        max_length=255,  # If we need future states
+        blank=True,
+        null=True,
+        help_text='Fetching Status',
+        choices=ProposalStatus.choices,
+    )
+    ground_truth = models.BooleanField(blank=True, null=True)
```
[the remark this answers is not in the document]
BryonLewis: Yeah, when I made it I allowed it to be null, but setting it more explicitly to False would probably better.

django/src/rdwatch/views/model_run.py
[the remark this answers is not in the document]
BryonLewis: I guess the plan would be:

1. endpoint which kicks of a task to generate a zip file for a model-run.  Returning the taskId for celery
2. Client polls the celery task-id to see if the task is finished.  Once finished it can ask for the file associated with the taskId
3. While running the task creates the zip file and stores it (not sure if I store it in S3 or on system)
4. The client once it knows the task is finished will request to download the file
5. After downloading the file it will remove it (either from system or S3).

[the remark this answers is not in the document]
BryonLewis: Looking into it again (I forgot the worker is in another container and could be in another machine without shared storage), the best/easiest option may be an 'AnnotationExports' table which utilizes S3 FileField to store the zip file.  It would contain the taskId, the model-run and a timestamp.  A third indicator will there to let it know if it's been downloaded and then we can use the schedule clean up task to remove downloaded files after an hour or so (whenever the scheduled clean up task runs).

BryonLewis:
I converted the downloading of zip files into a celery task.  For ease I created another table (AnnotationExports, I'm open to any vocabulary/naming changes).  When the task is complete it uploads a zip file to S3/MinIO and returns to the client a pre-signed URL for downloading. 
During the task process the client polls a new endpoint using the taskId to get the current status of the download process.

I added a new celery beat schedule to delete any annotation exports that are over an hour old.

django/src/rdwatch/models/site_evaluation.py
```diff
@@ -63,20 +105,50 @@ def bulk_create_from_site_model(
...
+        status = None
+        modified = False
+        if (
+            configuration.ground_truth is False
+            and site_feature.properties.validated
+            and site_feature.properties.originator == 'te'
+        ):
+            configuration.ground_truth = True
+            modified = True
+        if configuration.proposal or site_feature.properties.site_number == 9999:
```
mvandenburgh, line 117: Can you add a comment mentioning the significance of 9999?

django/src/rdwatch/tasks/__init__.py
```diff
@@ -295,3 +331,42 @@ def delete_temp_model_runs_task() -> None:
...
         model_runs_to_delete.delete()
+
+
+@shared_task
+def delete_export_files() -> None:
+    """Delete all S3 Export Files that are over an hour old."""
+    exports_to_delete = AnnotationExport.objects.filter(
+        created__lte=timezone.now() - timedelta(hours=1)
+    )
+    with transaction.atomic():
```
mvandenburgh, line 342:
```suggestion
```

django/src/rdwatch/tasks/__init__.py
```diff
@@ -295,3 +331,42 @@ def delete_temp_model_runs_task() -> None:
...
+def download_annotations(self, id: int):
+    # Needs to go through the siteEvaluations and download one for each file
+    site_evals = SiteEvaluation.objects.filter(configuration_id=id)
+    configuration = HyperParameters.objects.get(pk=id)
+    task_id = self.request.id
+    temp_zip_file = tempfile.NamedTemporaryFile(delete=False, suffix='.zip')
+
+    with tempfile.TemporaryDirectory() as temp_dir:
+        file_list = []
+        for item in site_evals:
```
mvandenburgh, line 356:
```suggestion
        for item in site_evals.iterator():
```

django/src/rdwatch/views/model_run.py
```diff
@@ -456,3 +515,32 @@ def get_modelrun_evaluations(request: HttpRequest, hyper_parameters_id: int):
         }
     query['region'] = model_run['region']
     return 200, query
+
+
+@router.post('/{id}/download')
```
mvandenburgh, line 520: The current convention elsewhere in the codebase is to use trailing slashes (applies to the other endpoints introduced in this PR too)
```suggestion
@router.post('/{id}/download/')
```

django/src/rdwatch/views/model_run.py
```diff
@@ -456,3 +515,32 @@ def get_modelrun_evaluations(request: HttpRequest, hyper_parameters_id: int):
...
+def check_download(request: HttpRequest, task_id: str):
+    task = AsyncResult(task_id)
+    print(task)
+    return task.status
+
+
+@router.get('/{id}/download')
+def get_downloaded_annotations(request: HttpRequest, id: int, task_id: str):
+    annotation_export = AnnotationExport.objects.filter(celery_id=task_id)
+    print(annotation_export)
```
mvandenburgh, line 537:
```suggestion
```

mvandenburgh (changes requested):

mvandenburgh (commented):
LGTM. Any ideas why the `check-migrations` check is failing?

BryonLewis:
UPDATE:  It was my mistake, I thought I had made migrations after my final change but I didn't.  I've updated the 0012 migrations to reflect all the changes now.

mvandenburgh (approved):

PR #253, opened 2023-08-22
Remove remaining DRF imports and uninstall it

mvandenburgh:
With #234, we are completely converted to ninja. This PR removes some remaining loose ends/DRF imports and uninstalls the dependencies from our poetry lockfile.


BryonLewis (approved):
Did a clean build and ran it locally.  Seems to work fine.

Just a note that I think we need to handle the GEOGeometry issue and we may be able to get /api/docs working properly.  That can be another PR though,

PR #254, opened 2023-08-23
DRAFT: Scoring endpoints apl

Sean-S-Wang:
# Draft PR for integrating T&E scoring database into RD Watch


Sean-S-Wang:
# Additional tasks
## Vector tiles display
Waiting on some updates to schemas to aid in rendering easier

## Create endpoints for cancelling and starting satellite fetching/images (and collection of model run (aka evaluation run downloads))

For satellite downloading, monitor the celery task to see if its working

## Proposal viewing
Need to find database where they store proposal sites, and extract proposals, group them so model run can show them and make them look like model run, and some additional endpoints used with proposal view

There is a patch endpoint (`patch.site_evaluation` allows to edit site evlauations). Currently have `SiteEvaluationTracking` patch. This has to be done in the rdwatch database as to not pollute the scoring database, but still be able to track changes long term. 

When editing, want to update in place (might not be allowed), or keep a log of it and edit in place.

Might need to put in additional table to track the state of the proposal. Already done for observations in rdwatch database for approve or reject status, but need to do for proposal.

Should create our own models and shape them how we want via a `models` folder in `rdwatch_scoring`.

## Ground truth updating
Need to figure out how to update data within the scoring database. Once that is done, just adding something to interface/view about approve/change, and that would update the scoring db itself. Right now, just editing `time range` and `label` of it, as well as adding notes (this might need to be a new column).

It does a filter when we get evaluations to see if any are proposals (Entire region, collection of sites, which are a collection of observations). When clicked, loads all sites on right side, select each of them, and then set accept/rejected/waiting state for each observation, then can export each geojson back out.

## Additional notes
Once site evaluation stuff is done, should be easier to implement downstream features.

Sean-S-Wang:
This can be closed?

PR #255, opened 2023-08-24
Update Poetry, add `werkzeug` as a dev dependency.

mvandenburgh:
We discovered today that `werkzeug` is required to run `runserver_plus`, but isn't included as a dev dependency. This PR adds it and also upgrades poetry while we're at it.


mvandenburgh:
cc @BryonLewis @Sean-S-Wang

PR #256, opened 2023-08-28
Remove observation searching for geometry  in model-run endpoint

BryonLewis:
I made a sample set by reloading the Dubai region multiple times to get a sense of where the performance issues were. 

I had a hunch that it was searching into each observation.  I had removed most observation searches by moving the date/time range up into the Evaluation area.

I decided to remove the bbox case of looking at observation geometries and focus more on the evaluation geometry.  I also could remove the observation count by removing this.  Now a model run doesn't have to look into each site observation (of which there could be hundreds per site evaluation) to calculate geometry, it will grab it from the higher level.

There could be an argument for the future that the model-run list doesn't need geometry at all and it could be removed in the future so that when a user clicks on a model-run it then grabs the geometry for that specific run.

Take like 3 seconds to get the info for 64 copies of the AE_R001 region
image

### DIFF
--- a/django/src/rdwatch/views/model_run.py
+++ b/django/src/rdwatch/views/model_run.py
@@ -174,7 +174,6 @@ def get_queryset():
         .order_by('-groundtruth', '-id')
         .alias(
             region_id=F('evaluations__region_id'),
-            observation_count=Count('evaluations__observations'),
             evaluation_configuration=F('evaluations__configuration'),
             proposal_val=F('proposal'),
         )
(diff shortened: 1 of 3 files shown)


mvandenburgh (approved):
Nice find! No objections from me, although I don't know what the downsides are of using the site evaluation geometry instead of the union of each observation geometry.

BryonLewis:
> Nice find! No objections from me, although I don't know what the downsides are of using the site evaluation geometry instead of the union of each observation geometry.

The only downside is that it may not zoom in properly if there is no site evaluation geometry.  However according to our discussion with Sean the other day, nothing in the Scoring DB would not have a site-level geometry.

PR #258, opened 2023-08-31
Update performer filter to allow multi-select

BryonLewis:
Per the goals document for the MVP one of the items listed was enabling multiple performers to be selected.

This PR updates the back-end and front-end to allow for multiple selections of performers.  Mostly just changes the 'performer' filter field to an a list of performers instead of a single performer.

The front-end right now increases the height of the box to fit more performers, we may want to revisit this behavior.


mvandenburgh (approved):
LGTM

PR #259, opened 2023-09-06
WV pan-sharpening

BryonLewis:
- Modifies the STAC Search to get pan images as well as RGB images when getting WorldView data.
- switches stored images to PNGs to reduce compression
- adds a bits_per_pixel to processed captures for future range scaling based on this information if it is found in the nitf.
- Raster_Tile will now use the pan.uri if it is found during the STAC search and create a pan-sharpened image.  there is some logging for download/poceessing times for debugging.  We may want to leave those in there so we can see in the logs if something gets stuck.
- All Vue changes were to get rid of warnings shown in PRs.


mvandenburgh (approved):
Tested locally and everything looks good :+1:

PR #261, opened 2023-09-07
MVP Cleanup

### LINKED ISSUE #260: MVP Final Modifications
- Allow for duplicate 9999 region Ids to be loaded and displayed in the system:
    - Use the cache_file originator as a display name for these duplicate Ids
    - Update the endpoint that gets a list of regions for proposals to include the cache originator filename if it exists
    - Display the Base Id and the cache originator file in the list.  The cache originator may need to be a truncated tooltip given that the file name could be quite long.
    - Update exporting of the files in the zip file to utilize the cache_originator name if it exists.
- Add Notes to the cache field on export for the site model.
- Add download settings to downloading of model-runs
    - All Annotations
    - Approve Annotations
    - Rejected Annotations
    - Just a secondary dialog when you click download to configure what to download.  This should only be necessary on proposals.  For everything else it should default to all.
- Add an option to turn off the text display to the settings.

BryonLewis:
resolves #260 

- Allows for multiple proposals to have the same siteId (KR_001_9999)
    - I add a `-#` to any subsequent repeating Ids for the user interface
    - The exporting in a Zip file is updated to export the JSON files with the -# usage.  I tried using the cache_originator field but there is nothing that indicates that the field is unique amongst the proposals.
    - This required removing the constraint on uniqueness for SiteEvaluation based on the configuration/region/number because we can have duplicate numbers now.
    - Added the capability to view the source file for site proposals as well.
 - Downloading Annotations Filtering
     - When using proposals for downloading a dialog will come up which allows the user to filter the downloads by 'all', 'approved' or 'rejected'.  The corresponding zip file will contain only those files whose status matches the filter.
 - The Text String is now hidden by default, and there are settings where the user can toggle the text string back on if they want.

image
image


mvandenburgh (approved):
LGTM, tested locally

PR #265, opened 2023-09-18
Use `pystac-client` to make STAC requests

mvandenburgh:
This PR installs `pystac-client` and uses it to query the STAC API instead of manually crafting HTTP requests. The next step is to remove the code that parses the raw JSON blob from the STAC server, and instead use the pystac-client's built in parsing functionality; I'll include that in a follow up PR. 

One additional thing to note is that the pystac client handles pagination
automatically. So, I've removed the manual pagination handling code as well.

### DIFF
--- a/django/src/rdwatch/utils/satellite_bands.py
+++ b/django/src/rdwatch/utils/satellite_bands.py
@@ -31,113 +31,96 @@ def get_bands(
     if constellation.slug not in ('L8', 'S2'):
         raise ValueError(f"Unsupported constellation '{constellation.slug}'")
 
-    page: None | int = 1
+    results = stac_search(
+        constellation.slug,
+        timestamp,
+        bbox,
+        timebuffer=timebuffer or timedelta(hours=1),
+    )
 
-    while page is not None:
-        results = stac_search(
-            constellation.slug,
-            timestamp,
-            bbox,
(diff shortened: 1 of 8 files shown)


django/src/rdwatch/utils/satellite_bands.py
```diff
@@ -31,113 +31,96 @@ def get_bands(
...
+        constellation.slug,
+        timestamp,
+        bbox,
+        timebuffer=timebuffer or timedelta(hours=1),
+    )
+
+    if 'features' not in results:
+        logger.warning("Malformed STAC response: no 'features'")
+        return
+
```
mvandenburgh, line 44: A note about this file - it looks like a lot has changed here, but all I did was remove the while loop and the conditional that checks the `links` field for pagination info. The large diff is due to the de-indent I had to do after removing the while loop.

BryonLewis (approved):
There was a poetry conflict (just the hash) so I resolved that.  Pulled and tested, even seems to work with the api.smart-stack.com  without having to change any of the inputs.

I'm going to merge this in and remove the `SearchParams` Class in my Pansharpening branch because they aren't used anymore.
https://github.com/ResonantGeoData/RD-WATCH/pull/265/files#diff-aca727bdd5046dbf5444c00fc128bef61be5d291edd4de57a90acd3aa502e8a6R45-R50

PR #266, opened 2023-09-21
Denormalize region name into a single column

mvandenburgh:
Currently in the RGD database, we break region strings up into multiple columns (i.e. for `AE_R001`, we store the country (AE), type (R), and id (0001) in separate columns). The T&E database on the other hand simply stores the region as a single string. This PR brings the RGD db in line with the T&E db.

This is a prerequisite for embedding region names directly in the vector tile metadata (at least, it makes it significantly simpler); I'll make a follow up PR that implements that change.

### DIFF
--- a/django/src/rdwatch/admin.py
+++ b/django/src/rdwatch/admin.py
@@ -41,12 +41,6 @@ class ProcessingLevelAdmin(admin.ModelAdmin):
     search_fields = ('slug',)
 
 
-@admin.register(lookups.RegionClassification)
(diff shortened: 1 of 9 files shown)


mvandenburgh:
@BryonLewis I tried to test this as extensively as I could, but feel free to play around with it and try to break it and see if I missed anything. One thing I'll note is that the `/api/evaluations` endpoint does not work and returns a 500, however that appears to be unrelated to this PR and seems to happen even on `main`.

BryonLewis (approved):
Pulled, tested and looks good.

PR #268, opened 2023-09-22
Refactor backend + frontend to reference `Regions` by string name instead of numerical ID

mvandenburgh:
This PR removes any notion of region "IDs" from the external-facing API. Regions are now referenced entirely by their string, just as in the T&E database.

### DIFF
--- a/vue/src/client/services/ApiService.ts
+++ b/vue/src/client/services/ApiService.ts
@@ -157,8 +157,8 @@ export class ApiService {
           }
         });
       }
-      
-  
+
+
       /**
    * @param id
    * @returns boolean
@@ -213,7 +213,7 @@ export class ApiService {
           },
         });
       }
-  
+
   /**
    * @returns ModelRunList
    * @throws ApiError
@@ -243,8 +243,8 @@ export class ApiService {
         },
       });
     }
-  
-    
+
+
 
   /**
    * @param id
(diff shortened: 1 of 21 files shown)


mvandenburgh:
@BryonLewis once again this is a large refactor unfortunately. The fact that typescript passes after changing the `Region` type to simply be a `string` on the frontend gives me some degree of confidence that I got this right, but let me know if I missed anything.

mvandenburgh:
Ok, i noticed a bug - selecting a region and then deselecting it causes `null` to be populated in the dropdown. Looking into this

mvandenburgh:
I found the issue, the fact that we're not using a `regionMap` anymore meant that the initialization of the region from the url string needed to be changed from a `watch` to `onMounted` (in `RGD.vue` and `AnnotationViewer.vue`) to avoid clobbering the value. Also, for some reason the `v-select` component seems to behave differently when a string is passed as a `v-model` - instead of setting it to `undefined` on clear, it sets it to `null`. 

While debugging this issue, I refactored the `queryFilters` variable by changing it from a `ref` to a `computed` value. Ultimately that didn't have any effect on the bug, but I kept it in this PR because I think it's an improvement - see https://github.com/ResonantGeoData/RD-WATCH/pull/268/commits/5acc1a0e704a323c2852680f0ff68984b1d7edf2.

vue/src/client/services/ApiService.ts
```diff
@@ -19,7 +19,6 @@ import { EvaluationImageResults } from "../../types";
 
 export interface QueryArguments {
   performer?: string[];
-  groundtruth?: boolean;
```
BryonLewis, line 22: I'm guessing this was removed because it isn't being used.  I might have added that before as a way to eventually only show the groundtruth but never implemented it in `views/model_run.py`

BryonLewis (changes requested):
I think once my PR is merged into this it seems like it should be good based on my simple testing and review.

BryonLewis (approved):
:+1:

PR #269, opened 2023-09-22
Migrate `HyperParameters` to `ModelRun`, use UUID as primary key

mvandenburgh:
This PR creates a new model, `ModelRun`, that is identical to `HyperParameters` except it uses UUIDs for its primary keys instead of integers. It doesn't seem possible to change the primary key type for an existing Django table without writing manual SQL queries outside of Django's migrations system, so I thought I would take this opportunity to do this rename as well.


django/src/rdwatch/models/model_run.py
```diff
@@ -1,9 +1,13 @@
...
 from django.db import models
 
 
-class HyperParameters(models.Model):
+class ModelRun(models.Model):
+    id = models.UUIDField(primary_key=True, default=uuid4, editable=False)
+
     created = CreationDateTimeField()
     title = models.CharField(max_length=1000)
     performer = models.ForeignKey(
```
BryonLewis, line 13: What's your opinion on adding a regions column that contains an array of region names?  We could update it when sites are posted to a model-run.  Then our filtering could run directly on the model-run table and only after we get the first X number of model runs could we calculate the other values required (Number of sites, average score others).

mvandenburgh: I think duplicating that info would not be ideal, since it would ruin the referential integrity of our schema. I think adding caching like we have in the vector tile endpoint would actually be preferred to that, since that comes with the same tradeoffs (namely, having to deal with invalidation/updating derived values whenever we update the source data), but without touching the database/anything persistent. 

But more generally, I agree something like this would be useful. Maybe we could consider a [stored generated column](https://www.postgresql.org/docs/current/ddl-generated-columns.html).

BryonLewis (approved):
I pulled and tested and it looks good.  Thanks for doing this.

PR #272, opened 2023-09-25
Migrate site evaluations + observations to use UUIDs instead of integer ids

mvandenburgh:


BryonLewis (changes requested):
Most everything works except for the reverse hovering when hovering over a card.  I think we may just disable that feature and try to resolve it in the future or re-write it.

BryonLewis (approved):

PR #273, opened 2023-09-26
Fixes the site image endpoints by changing the int to str

BryonLewis:
Just noticed two locations where the siteobs_id was still an int which would cause nasty ninja validation errors when downloading images or viewing them.

There is a secondary problem that I can't figure out at all.  If you bring up a model run and click on a polygon it displays in the right side of the screen.  If you hover over that card on the right side of the screen it's supposed to highlight the polygon by making it thicker.  This is to associate your card with a polygon.  For some reason this makes all annotations disappear at once when you hover over the card.  I tried looking into the filters and a bunch of areas but this has me stumped.

### DIFF
--- a/django/src/rdwatch/views/site_image.py
+++ b/django/src/rdwatch/views/site_image.py
@@ -20,7 +20,7 @@ class SiteImageSchema(Schema):
     source: str
     cloudcover: float
     image: str
-    siteobs_id: int | None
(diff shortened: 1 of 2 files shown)


mvandenburgh (approved):
Hmm, I'll try to look into that other problem

PR #275, opened 2023-09-27
Fix docker image build

mvandenburgh:
`unit` no longer supports Ubuntu 22.10 (see https://unit.nginx.org/installation/#ubuntu-2210), which appears to be the cause of the docker build errors on `main`. This PR upgrades the RD-WATCH docker image to Ubuntu 23.04. This has a consequence of requiring Python 3.11, as `unit` requires it on Ubuntu 23.04.


django/pyproject.toml
```diff
@@ -25,13 +25,13 @@ packages = [
...
 [tool.poetry.dependencies]
-python = "~3.10.0"
+python = "~3.11.0"
 django = ">=4.1,<4.2"
 psycopg2 = "~2.9.3"
 redis = { version = "~4.3.4", extras = [ "hiredis" ] }
 uritemplate = "^4.1.1"
 iso3166 = "^2.1.1"
-rio-tiler = "^3.1.6"
+rio-tiler = "5.0.3"
```
mvandenburgh, line 34: Note, this one package had t be updated to be installable on Python 3.11. We should do an upgrade of the rest our python dependencies at some point.

mvandenburgh:
Looking into test failures...

BryonLewis (approved):
I pulled, built it and did some basic testing and it looks like everything is functioning properly.

PR #276, opened 2023-09-27
Remove migrations/fixture loading from docker entrypoint

mvandenburgh:
Running these on **every** container start is causing issues on our ECS deployment on AWS. Specifically, the problem is that the migrations in #272 take several minutes to run; on top of that, each container (the django web app, celery worker, and celery beat) each tries to run the migration command at once. This results in one of them acquiring a lock and beginning the migrations, while the other two sit there waiting. While not ideal, this wasn't an issue for short-running migrations up to this point. But now that we have a minutes-long migration, it looks like some Django mechanism that detects deadlocks kicks in for the other two containers and causes the container to exit with a non-zero status code. This in turn signals to CloudFormation that the deploy failed, which causes it to roll back to the previous image.

I think there's also some valid reasons for not wanting this behavior in local development either; for example, if you're writing a new migration, you may not want to have migrations automatically applied by docker. 

Instead, I propose we remove the migrations (and the associated `loaddata` command) from the docker entrypoint, and instead add them to only _one_ of the container entrypoints in the ECS container definition. And for local development, developers can run `manage.py migrate` themselves when they wish to run migrations.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -9,16 +9,18 @@ This is the simplest configuration for developers to start with.
 2. Set the environment variables in `.env`.
 3. Run `docker compose up` to start the Django development server and Celery worker, plus all backing services
    like PostGIS, Redis, RabbitMQ, etc.
-4. Optionally, populate your database with test data by running `docker compose run --rm django poetry run django-admin loaddata testdata`
-5. Optionally, create an account for the Django admin (http://localhost:8000/admin) by running `docker compose run --rm django poetry --directory django run django-admin createsuperuser`
-6. Start the client development server:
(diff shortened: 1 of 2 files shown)


mvandenburgh:
@BryonLewis if you agree with these changes, I'll make the necessary updates in the infrastructure repo before merging this.

BryonLewis:
I agree with this.  I think I mentioned a while ago removing the auto migrations because it makes switching between stuff sometimes a pain with it failing migrations.

BryonLewis (approved):
I know we are getting away from lookups.  But we may want to document some of this stuff in the Development.md.

mvandenburgh:
> I know we are getting away from lookups. But we may want to document some of this stuff in the Development.md.

Thanks, I updated the Development.md with the new state of things.

PR #277, opened 2023-09-28
Downloading Image Sequences

BryonLewis:
- Adds a tool called 'canvas-capture' which is based on 'ccapture.js' for creating GIFs in the client and downloading them from a canvas.  That's why there are package.json and package-lock.json changes.
- On hitting the download button a secondary hidden canvas is created and draws the images uses requestAnimationFrame, the canvas-capture tool then records the information and generates a GIF once complete.
- The drawing tool for the images has been modified to accept an override Width/Height which can resize to a specific value.  This is used to find the max size of the images in the `filteredImages` list and then rescale everything to the same size.  If all images are below 500px in any dimension it will resize them so that the smallest dimension is 500px.
- Added a section to the draw images to create a legend that would include the Source, Date, Activity.
- For right now there is a default quality and default FPS of 1Hz.  We can update these in the future when needed.
- Progress indicator is used to indicate the drawing section of the process and then the GIF creation
- Settings option is used to set the FPS and quality levels (I don't think quality matters much with full screen changes but added it anyways.
- Refactored some bigger functions in the Vue component into an `imageUtils.ts` file.
- If you see a console error just turn of your vue development tools.  The Vue error doesn't seem to effect any rendering/functionality and is only there in the development tools.  This seems to be a known issue with a dependency that all tools (React/Vue) seem to be using a similar global name.

### DIFF
--- a/vue/package-lock.json
+++ b/vue/package-lock.json
@@ -8,6 +8,7 @@
       "dependencies": {
         "@fontsource/inter": "^4.5.12",
         "@fontsource/iosevka": "^4.5.4",
+        "canvas-capture": "^2.1.1",
         "lodash": "^4.17.21",
         "maplibre-gl": "^2.1.9",
         "vue": "^3.2.37",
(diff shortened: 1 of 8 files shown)


mvandenburgh:
> Progress indicator is used to indicate the drawing section of the process and then the GIF creation

Is there supposed to be some kind of loading indicator when the GIF is being created? When I try this locally, after clicking the create gif button, nothing happens for about 20 seconds and then my browser starts downloading the gif.

BryonLewis:
> > Progress indicator is used to indicate the drawing section of the process and then the GIF creation
> 
> Is there supposed to be some kind of loading indicator when the GIF is being created? When I try this locally, after clicking the create gif button, nothing happens for about 20 seconds and then my browser starts downloading the gif.

Maybe there is a CSS issue, but there should be a loading bar and state information:

https://github.com/ResonantGeoData/RD-WATCH/assets/61746913/1e4d45cc-c7af-464c-999d-d218de8f7671

mvandenburgh (approved):
Ok, I tried it in an incognito window (where I have vue dev tools completely disabled) and it works now. Not sure if I didn't disable vue dev tools correctly before, or if the merge from `main` fixed it. Either way it works as expected for me :+1:

PR #278, opened 2023-09-28
pan sharpening fix

BryonLewis:
- I had to previously change some `rgb.data =` to `rgb.from_array(` but I missed assignment before.  I updated that so the pan-sharpening images should now download properly.
- While I was in here I added a download option to use the bits per pixel to scale the image instead of the default value of 10,000.  This will brighten up a lot of the dark images.
- I then added the corresponding UI to the front-end when selecting WV images to show that option for changing the scaling method from 'default' to 'bits'.

### DIFF
--- a/django/src/rdwatch/tasks/__init__.py
+++ b/django/src/rdwatch/tasks/__init__.py
@@ -69,6 +69,7 @@ def get_siteobservations_images(
     dayRange=14,
     no_data_limit=50,
     overrideDates: None | list[datetime, datetime] = None,
+    scale='default',
 ) -> None:
     constellationObj = Constellation.objects.filter(slug=baseConstellation).first()
     # Ensure we are using ints for the DayRange and no_data_limit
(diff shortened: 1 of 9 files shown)


django/src/rdwatch/utils/worldview_processed/raster_tile.py
```diff
@@ -54,6 +56,7 @@ def get_worldview_processed_visual_bbox(
     capture: WorldViewProcessedCapture,
     bbox: tuple[float, float, float, float],
     format='PNG',
+    scale='default',
```
mvandenburgh, line 59:
```suggestion
    scale: Literal['default', 'bits'] = 'default',
```

django/src/rdwatch/utils/images.py
```diff
@@ -90,6 +90,7 @@ def fetch_boundbox_image(
     timestamp: datetime,
     constellation: str,
     worldView=False,
+    scale='default',
```
mvandenburgh, line 93:
```suggestion
    scale: Literal['default', 'bits'] = 'default'
```

django/src/rdwatch/tasks/__init__.py
```diff
@@ -69,6 +69,7 @@ def get_siteobservations_images(
     dayRange=14,
     no_data_limit=50,
     overrideDates: None | list[datetime, datetime] = None,
+    scale='default',
```
mvandenburgh, line 72:
```suggestion
    scale: Literal['default', 'bits'] = 'default'
```

mvandenburgh (approved):
Just some suggested type hints to improve code documentation. LGTM and works for me locally.

PR #279, opened 2023-09-29
Update model run upload script for new endpoint response format

mvandenburgh:
Regions are now returned from the API as simple strings, not objects. I missed this in #268, and noticed this when uploading proposals to test #278.


BryonLewis (approved):
I forgot to mention this, I did the same in my local version.  I keep a copy of the scripts and a bunch of annotation folders (GT, scoring, Proposals) in a separate folder so I can quickly repopulate the DB.  Since the edits were in another folder it wasn't in my git changes.

PR #281, opened 2023-10-03
Fix ninja swagger page not working

mvandenburgh:
The autogenerated swagger page provided by ninja wasn't working due to the use of complex types like `GEOSGeometry` and `Polygon` directly in our pydantic request models. I moved those to dynamic properties and updated the ingest code to use those, and it works now. See `/api/docs/`.


BryonLewis (approved):
:+1:

PR #282, opened 2023-10-04
Switch to ninja's built-in pagination for model runs endpoint/eliminate JSONB subquery

mvandenburgh:
This PR switches the model run endpoint to use ninja's built-in pagination instead of a manual approach; this involved refactoring the query to remove the JSONB aggregation, which improves readability regardless. Note I haven't removed any fields yet, and the structure of the response returned by the model run LIST and GET endpoints remains mostly unchanged. The exception to this is that `results` becomes `items` in the list endpoint response; this was done to bring the endpoint in line with ninja's default pagination behavior. But in general, this PR is just a refactor that will make it easier to do more extensive changes to the endpoint in a followup PR. 

However, removing the large JSONB aggregation does seems to improve performance on its own, bringing the response time down from about 5 seconds to about 3.5 seconds for me locally with 361 model runs.

### DIFF
--- a/django/src/rdwatch/views/model_run.py
+++ b/django/src/rdwatch/views/model_run.py
@@ -129,6 +129,28 @@ class ModelRunListSchema(Schema):
 
 
 def get_queryset():
+    # Subquery to count unique SiteEvaluations
+    # with proposal='PROPOSAL' for each ModelRun
+    proposed_count_subquery = (
+        SiteEvaluation.objects.filter(
+            configuration=OuterRef('pk'),
+            status='PROPOSAL',
+        )
+        .values('configuration')
+        .annotate(count=Count('pk'))
+        .values('count')
+    )
+
+    # Subquery to count unique SiteEvaluations with proposal
+    # not equal to 'PROPOSAL' for each ModelRun
+    other_count_subquery = (
+        SiteEvaluation.objects.filter(configuration=OuterRef('pk'))
+        .exclude(status='PROPOSAL')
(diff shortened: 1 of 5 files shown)


django/src/rdwatch/views/model_run.py
```diff
@@ -163,67 +151,75 @@ def get_queryset():
...
-                proposal='proposal_val',
-                adjudicated=Case(
-                    When(
-                        ~Q(proposal_val=None),  # When proposal has a value
-                        then=JSONObject(
-                            proposed=Coalesce(Subquery(proposed_count_subquery), 0),
-                            other=Coalesce(Subquery(other_count_subquery), 0),
-                        ),
-                    ),
-                    default=None,
```
BryonLewis, line 220: Any way we can keep this.  It's a simple thing that should only be run when the proposal_val is non-null.  It is used for the annotators to see the current status of a model run and indicate the proposed and remaining sites?

They don't need to be in the JSONB object.

BryonLewis (changes requested):
just curious if the proposal query is too big of impact?  It's just checking if the proposal field exists and during a secondary query if it is.

django/src/rdwatch/views/model_run.py
mvandenburgh: My bad, somehow I missed that. The intent of this PR is to keep all the original fields, so I added that back in.

BryonLewis (approved):
LGTM

PR #286, opened 2023-10-10
Ignore unset values in PATCH request body

mvandenburgh:
Calling `.dict()` on a pydantic model instance with `exclude_unset` set will cause values that are not present/undefined to simply be excluded from the resulting dict instead of casting them to `None`, while values that are explicitly set to `null` in the request body will be cast to `None` as expected. 

[Docs for `exclude_unset`](https://docs.pydantic.dev/1.10/usage/exporting_models/#modeldict)


BryonLewis (approved):
Thanks for finding that.

PR #290, opened 2023-10-12
Log response body if model run ingest fails

mvandenburgh:
This logs out the response body if an ingest error occurs. Currently, it just reports a `422` error.

T&E was seeing some 422 errors coming back from RGD due to a validation error with the site models they were uploading using this script, but it wasn't clear what the problem was because nothing was being logged.

### DIFF
--- a/scripts/loadModelRun.py
+++ b/scripts/loadModelRun.py
@@ -7,6 +7,7 @@ import sys
 import traceback
 from glob import glob
 from multiprocessing import Pool
+from pathlib import Path
 
 import requests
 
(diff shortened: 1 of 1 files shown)


scripts/loadModelRun.py
```diff
@@ -160,7 +161,13 @@ def post_site(post_site_url, site_filepath, rgd_auth_cookie):
...
     print(response)
     if response.status_code != 201:
-        print(f'Error uploading site, status ' f'code: [{response.status_code}]')
+        print(
+            f'Error uploading site "{Path(site_filepath).name}", status '
+            f'code: [{response.status_code}]',
+            file=sys.stderr,
+        )
+        print(response.text, file=sys.stderr)
+    print()
```
BryonLewis, line 170: was this intentional to print a new line?

BryonLewis (approved):

PR #292, opened 2023-10-12
bbox min size, fix counting, start custom bit scaling

### LINKED ISSUE #285: Add Satellite Image Downloading Options
- Ensure it is using the siteExtents and not the siteObs bounding box for getting a subset of an image
- Allow custom pixel scaling range to be set as a parameter.  The 0-10000 range is the default and we have the bit level scaling as well.  We can add a custom setting to specify the scaling range.  I'm guessing we set a default 0-10000, and have a bit scaling (which uses the bit level of the image).
- Fix the count for site-obs downloads, it is using the count of all sitesObs instead of the count of the site-obs that match the type.
- Create a minimum area for image downloads in S2/L8.  Some models are too small to gain much information, This could add more area around the bounding box for display purposes.  So at a minimum S2/L8 downloads are X size.
- Add a custom bounding box addition to the downloads.  By default, it uses 20% which would allow increasing this bounds to make it larger.

BryonLewis:
resolves #285 

- Resolved the counting issues when downloading sites
- Made it so there is a custom size for minimum S2/L8 images, currently set to have at least a 500m width (estimate).  If it's under that it will add 250m to each existing dimension.  This required some front-end changes to realign the image properly.
- There is now a choice for custom scaling of 'default (0-10000), 'bits (which will use image information to scale', and 'custom (which takes input from the user for the images).
- Added a bbox scaling option to scale the additional size area around the bounding box.  The default is still 1.2x the bounding box size.
- Form validation for the bbox and scale values when downloading.

There were a bunch of front-end changes to the drawing of images/polygons to handle the scaling.  I'm going to list some of the modifications here:

- Images will at least download an image that 1.2x the base bounding box of the image.
- For S2/L8 images it will add extra space if the image is too small so the bounding box in any instance can be much bigger than 1.2x.
- When transferring lat/long coordinates to pixel space coordinates I now create a secondary polygon under the `PixelPoly` type called `scaled`.  This will automatically scale the base bounding-box of 1 to 1.2 times and check if it is significantly different from the current bounding box (~7% difference).  This difference is calculated because of all the conversions things aren't exactly perfect.
- Once the new bounds are created I calculate the amount of image to crop out to display the image.
- Next The polygon needs to be scaled relative to the image coordinates.

https://github.com/ResonantGeoData/RD-WATCH/assets/61746913/6c213a31-35ff-4ede-9e7d-91396cdc841d


BryonLewis:
TODO:

- [ ] What to do about different scaling sizes for different images and how to generate the GIF if are using different sizes.

Before the standard size was 20% or 1.2X the bounding box limits.  Now that the user can specify these custom and that smaller areas will have larger images we need to rescale these items.

Ideas:
- minimum size is 20% bounding box, and I know the standard bbox size so I could crop out any extra pixels.  I may have to store additional information in SatelliteImage for this to work properly.

mvandenburgh:
I get this error in my Django server logs when I try to download all images for a model run -

```
get_site_observation_images() takes from 2 to 3 positional arguments but 8 were given
  File "/home/local/KHQ/michael.vandenburgh/.cache/pypoetry/virtualenvs/rdwatch-nKCMUNfY-py3.11/lib/python3.11/site-packages/ninja/operation.py", line 104, in run
    result = self.view_func(request, **values)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/local/KHQ/michael.vandenburgh/Desktop/smart/RD-WATCH/django/src/rdwatch/views/model_run.py", line 373, in generate_images
    get_site_observation_images(
TypeError: get_site_observation_images() takes from 2 to 3 positional arguments but 8 were given
Internal Server Error: /api/model-runs/186e8589-841c-43cf-9924-890744207fcd/generate-images/
[25/Oct/2023 12:53:30] "POST /api/model-runs/186e8589-841c-43cf-9924-890744207fcd/generate-images/?constellation=WV&dayRange=14&noData=50&force=false&scale=default&bboxScale=1.2 HTTP/1.1" 500 558
```

Do you see that as well?

BryonLewis:
> I get this error in my Django server logs when I try to download all images for a model run -
> 
> ```
> get_site_observation_images() takes from 2 to 3 positional arguments but 8 were given
> Traceback (most recent call last):
>   File "/home/local/KHQ/michael.vandenburgh/.cache/pypoetry/virtualenvs/rdwatch-nKCMUNfY-py3.11/lib/python3.11/site-packages/ninja/operation.py", line 104, in run
>     result = self.view_func(request, **values)
>              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/local/KHQ/michael.vandenburgh/Desktop/smart/RD-WATCH/django/src/rdwatch/views/model_run.py", line 373, in generate_images
>     get_site_observation_images(
> TypeError: get_site_observation_images() takes from 2 to 3 positional arguments but 8 were given
> Internal Server Error: /api/model-runs/186e8589-841c-43cf-9924-890744207fcd/generate-images/
> [25/Oct/2023 12:53:30] "POST /api/model-runs/186e8589-841c-43cf-9924-890744207fcd/generate-images/?constellation=WV&dayRange=14&noData=50&force=false&scale=default&bboxScale=1.2 HTTP/1.1" 500 558
> ```
> 
> Do you see that as well?

Yes Sorry about that, I didn't test it as much downloading full sites and I changed the function to use a single argument.  I've just pushed a change that should fix it.

django/celerybeat-schedule.db
mvandenburgh, line 1: We should probably add this to the `.gitignore`.

django/src/rdwatch/views/site_observation.py
```diff
@@ -167,16 +170,22 @@ def site_observations(request: HttpRequest, evaluation_id: UUID4):
...
 
+class GenerateImagesSchema(Schema):
+    constellation: Literal['WV', 'S2', 'L8'] = ('WV',)
+    dayRange: int = (14,)
+    noData: int = (50,)
+    overrideDates: None | list[str] = (None,)
+    force: bool = (False,)
+    scale: Literal['default', 'bits', 'custom'] = ('default',)
+    scaleNum: None | list[int] = (None,)
+    bboxScale: None | float = (1.2,)
```
mvandenburgh, line 181: Why do these default to tuples? As they are, the default values don't adhere to the type hints.

```suggestion
    constellation: Literal['WV', 'S2', 'L8'] = 'WV'
    dayRange: int = 14
    noData: int = 50
    overrideDates: None | list[str] = None
    force: bool = False
    scale: Literal['default', 'bits', 'custom'] = 'default'
    scaleNum: None | list[int] = None
    bboxScale: None | float = 1.2
```

django/src/rdwatch/utils/raster_tile.py
```diff
@@ -23,17 +24,44 @@ def get_raster_tile(uri: str, z: int, x: int, y: int) -> bytes:
...
+                    elif scale == 'bits':
+                        stats = cog.statistics()
+                        low = 0
+                        high = 10000
+                        if 'b1' in stats.keys():
+                            stats_json = stats['b1']
+                            low = stats_json['percentile_2']
+                            high = stats_json['percentile_98']
+                        img.rescale(in_range=((low, high),))
+                    elif len(scale) == 2:
```
mvandenburgh, line 49: Maybe this is overly paranoid, but in theory you could pass a string of length 2 here and it would also fulfill this condition.
```suggestion
                    elif isinstance(scale, list) and len(scale) == 2:
```

django/src/rdwatch/utils/raster_tile.py
```diff
@@ -23,17 +24,44 @@ def get_raster_tile(uri: str, z: int, x: int, y: int) -> bytes:
...
+            elif scale == 'bits':
+                stats = cog.statistics()
+                low = 0
+                high = 10000
+                if 'b1' in stats.keys():
+                    stats_json = stats['b1']
+                    low = stats_json['percentile_2']
+                    high = stats_json['percentile_98']
+                img.rescale(in_range=((low, high),))
+            elif len(scale) == 2:
```
mvandenburgh, line 65:
```suggestion
            elif isinstance(scale, list) and len(scale) == 2:
```
same as above

django/src/rdwatch/utils/worldview_processed/raster_tile.py
```diff
@@ -98,5 +98,7 @@ def get_worldview_processed_visual_bbox(
             if capture.bits_per_pixel != 8:
                 max_bits = 2**capture.bits_per_pixel - 1
                 rgb.rescale(in_range=((0, max_bits),))
+        elif len(scale) == 2:  # scale is an integeter range
```
mvandenburgh, line 101: I think you need to update the type hint for this in the function args. Also, same suggestion as above -
```suggestion
        elif isinstance(scale, list) and len(scale) == 2:  # scale is an integeter range
```

mvandenburgh (changes requested):

django/celerybeat-schedule.db
BryonLewis: I did last week, did I mess something up? https://github.com/ResonantGeoData/RD-WATCH/pull/295/files#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947

django/src/rdwatch/views/site_observation.py
BryonLewis: I feel like that was done by a formatter (and not our standard on) not me for some reason.  It doesn't make sense that I would have done that.

BryonLewis: I don't see anything in the commit history, guess I was just having a weird moment.

mvandenburgh (approved):

PR #293, opened 2023-10-16
Fix bug in ground truth upload script

mvandenburgh:
The `model_runs` dict wasn't being reset after site-models are finished uploading, causing all site-models to be reuploaded as region-models.


BryonLewis (approved):
Thanks, I typically skipped region models so I never noticed this.

PR #295, opened 2023-10-17
fixing annotation group export

BryonLewis:
I forgot to check the downloading of group annotations in our conversion to the UUID format.
This resulted in not being able to download groups of proposals.  I'd have to look more into why the type system didn't yell about this before and started yelling recently.

- Updated the relevant model-run endpoint to use str.  I don't know why I had it using int before, the task-id is stored as a charfield.
- Update the function for exporting annotations to use UUID4
- I added the celerybeat and noup.out to the .gitignore.  Was sick of having to skip those files when adding all my commit files.

### DIFF
--- a/django/src/rdwatch/views/model_run.py
+++ b/django/src/rdwatch/views/model_run.py
@@ -495,7 +495,7 @@ def check_download(request: HttpRequest, task_id: str):
 
 
 @router.get('/{id}/download')
-def get_downloaded_annotations(request: HttpRequest, id: int, task_id: str):
+def get_downloaded_annotations(request: HttpRequest, id: str, task_id: str):
     annotation_export = AnnotationExport.objects.filter(celery_id=task_id)
     if annotation_export.exists():
         annotation_export = annotation_export.first()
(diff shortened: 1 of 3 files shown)


django/src/rdwatch/views/model_run.py
```diff
@@ -495,7 +495,7 @@ def check_download(request: HttpRequest, task_id: str):
 
 
 @router.get('/{id}/download')
-def get_downloaded_annotations(request: HttpRequest, id: int, task_id: str):
+def get_downloaded_annotations(request: HttpRequest, id: str, task_id: str):
```
mvandenburgh, line 498:
```suggestion
def get_downloaded_annotations(request: HttpRequest, id: UUID4, task_id: str):
```

BryonLewis: I had that initially but noticed that the `AnnotationExport` model had a charfield even though it is a UUID.  Should we update that model to a uuid instead of a str?

mvandenburgh: Oh, I didn't realize. That `id` isn't even used, it's only the `task_id` is (which is indeed a `str` and not `UUID4`). Should we just remove it?

mvandenburgh (approved):

PR #299, opened 2023-10-24
Initial T&E Scoring Database Connection

mvandenburgh:
This PR is a slimmed down version of some changes from #254, in addition to some changes/improvements I pushed.

The first commit contains some changes cherrypicked from #254, including:
- Non-managed Django models generated from the T&E scoring database schema
- Implementations of the model run and vector tile endpoints for the scoring database
- Instrumentation of the vue-router in the client in order to allow switching between the RGD database and scoring database.

The rest of the commits contain the fixes and suggested improvements I made; the commit messages for each should be self-explanatory.


Sean-S-Wang:
I think all the functionality that we were aiming for is working, that being:
* EvaluationRun querying and loading in the left bar
* Clicking an evaluation run zooms into the region associated and clicking again brings up the observations associated with the site in the same color scheme as rdwatch
* Filtering by performer and region both are working as expected
* Adjusting the time slider correctly shows/hides observations depending on the timestamp attribute

django/src/rdwatch_scoring/views/vector_tile.py
```diff
@@ -0,0 +1,234 @@
...
+                configuration_id=F('evaluation_run_uuid'),
+                label=Case(
+                    When(
+                        Q(status_annotated__isnull=False),
+                        Lower(Replace('status_annotated', Value(' '), Value('_'))),
+                    ),
+                    default=Value('unknown'),
+                ),  # This needs a version to be scoring coloring,
+                # but that needs some coordination with kitware
+                timestamp=ExtractEpoch('evaluation_run_uuid__start_datetime'),
```
BryonLewis, line 108:
```suggestion
                timestamp=ExtractEpoch('start_date'),
```
This may be removed as well in our vector-tile endpoint I don't know if it's used but it is a timestamp which I think is a created timestamp.  It has no relation to any data I believe in the Evaluation.  Please double check this.

django/src/rdwatch_scoring/views/vector_tile.py
```diff
@@ -0,0 +1,234 @@
...
+        observations_queryset = (
+            Observation.objects.filter(
+                site_uuid__evaluation_run_uuid=evaluation_run_uuid
+            )
+            .alias(geomfromtext=geomfromobservationtext)
+            .alias(transformedgeom=transform)
+            .filter(intersects)
+            .values()
+            .annotate(
+                id=F('uuid'),
```
BryonLewis, line 137:
```suggestion
                id=F('uuid'),
                timestamp=F('date'),
```

So we in our version of the Observation model have a value called timestamp.  This is returned and provides the timestamp on hover.  That's why it isn't available on this system.  They have the 'date' property.  We just need to rename it to timestamp for it to show up.

image

django/src/rdwatch_scoring/views/vector_tile.py
```diff
@@ -0,0 +1,234 @@
...
+                mvtgeom=mvtgeom,
+                configuration_id=F('site_uuid__evaluation_run_uuid'),
+                site_number=F('site_uuid__site_id'),
```
BryonLewis, line 140:
```suggestion
                site_number=Substr(F('site_uuid__site_id'), 9),  # pos is 1 indexed
```
The site_number was only meant to be the number instead it is the whole region+number+performer I.E: 'KR_R001_0012_te_2.0.1'
Using the [Substr](https://docs.djangoproject.com/en/4.2/ref/models/database-functions/#substr) function and starting at position 9 should work.
image
image

This will also require importing Substr `from django.db.models.functions`

django/src/rdwatch_scoring/views/vector_tile.py
```diff
@@ -0,0 +1,234 @@
...
+                label=Case(
+                    When(
+                        Q(phase__isnull=False),
+                        Lower(Replace('phase', Value(' '), Value('_'))),
+                    ),
+                    default=Value('unknown'),
+                ),  # This should be an ID, on client side can make it understand this
+                area=Area(Transform('transformedgeom', srid=6933)),
```
BryonLewis, line 148: Within here we will also need:
```suggestion
                siteeval_id=F('site_uuid'),
```
This is used for clicking on a site and getting the details of it.  

We don't seem to have a site_evaluation details.  I.E `/api/scoring/observations/{uuid}`.  I don't think it's needed for this PR, but it will be required for getting images to download.

BryonLewis (changes requested):
This isn't a fully complete review but I wanted to give back some earlier feedback.

Mostly there are a few items in the vector tiles that were implicitly loaded from .values() I believe in our version and those fields needed to change/update when in the scoring DB.

We do need to implement probably in another PR the `/scoring/observation/{uuid}`.  This gets all the information about the downloaded data.

django/src/rdwatch_scoring/models/observation.py
```diff
@@ -0,0 +1,20 @@
...
+    sensor = models.CharField(max_length=255, blank=True, null=True)
+    phase = models.CharField(max_length=255, blank=True, null=True)
+    score = models.FloatField(blank=True, null=True)
+    crs = models.CharField(max_length=255, blank=True, null=True)
+    geometry = models.TextField(blank=True, null=True)
+    is_site_boundary = models.CharField(max_length=255, blank=True, null=True)
+    area = models.FloatField(blank=True, null=True)
+    confidence_score = models.FloatField(blank=True, null=True)
+
+    class Meta:
```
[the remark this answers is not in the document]
mvandenburgh: Because the model is defined inside the `rdwatch_scoring` app, Django implicitly uses that database. https://docs.djangoproject.com/en/4.2/ref/models/options/#app-label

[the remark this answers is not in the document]
BryonLewis: Thanks for the explanation.

django/src/rdwatch_scoring/views/vector_tile.py
mvandenburgh: I can implement that endpoint as part of the satellite image work.

django/src/rdwatch_scoring/views/vector_tile.py
mvandenburgh: Yeah, it's a created timestamp, this makes sense to me.

django/src/rdwatch_scoring/views/vector_tile.py
```diff
@@ -135,17 +135,18 @@ def vector_tile(
...
-                site_number=F('site_uuid__site_id'),
+                site_number=Substr(F('site_uuid__site_id'), 9),  # pos is 1 indexed
                 label=Case(
                     When(
                         Q(phase__isnull=False),
                         Lower(Replace('phase', Value(' '), Value('_'))),
                     ),
                     default=Value('unknown'),
                 ),  # This should be an ID, on client side can make it understand this
-                area=Area(Transform('transformedgeom', srid=6933)),
```
BryonLewis: This was my bad GitHub suggestion.  Didn't mean to overwrite this, this should be left in.

BryonLewis (changes requested):
Sorry, my bad with a github suggestion.

BryonLewis (approved):
👍

PR #301, opened 2023-10-26
fixing date/time picker and updating vuetify

BryonLewis:
Vuetify released an updated version with some changes to the Vue-Date-Picker: [v3.3.4](https://vuetifyjs.com/en/getting-started/release-notes/?version=v3.3.4
)
Some of the usage before has changed.

- I believe that before the model-value/v-model updated when the date was changed.  This one the value only updates when the user clicks 'OK'.
- I don't know if I really like this behavior but I needed to adjust the current component to fix it.
- I also added the ability for it to load the proper current date when initially clicking the edit button.

### DIFF
--- a/vue/package.json
+++ b/vue/package.json
@@ -24,7 +24,7 @@
     "maplibre-gl": "^2.1.9",
     "vue": "^3.2.37",
     "vue-router": "^4.2.4",
-    "vuetify": "^3.3.1"
+    "vuetify": "^3.3.4"
   },
   "devDependencies": {
     "@mdi/font": "^7.2.96",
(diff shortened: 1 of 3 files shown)


vue/package.json
```diff
@@ -24,7 +24,7 @@
     "maplibre-gl": "^2.1.9",
     "vue": "^3.2.37",
     "vue-router": "^4.2.4",
-    "vuetify": "^3.3.1"
+    "vuetify": "^3.3.4"
```
mvandenburgh, line 27: I think an update to `package-lock.json` is needed as well.

mvandenburgh (approved):

PR #303, opened 2023-10-26
Rename `SiteImage` fields to be consistent with scoring DB

mvandenburgh:


BryonLewis (approved):
Pulled, built did some basic downloading and image viewing and it seems to still work.

PR #309, opened 2023-11-07
Fix Scoring Satellite Fetching

BryonLewis:
Implements proper download counts for the /model-run and /model-run/{id} endpoints by adding in the model_run_uuid to the satellite fetching model for the scoring app and adding a secondary query to display the value.


django/src/rdwatch_scoring/models/satellite_fetching.py
```diff
@@ -5,3 +5,4 @@
 
 class SatelliteFetching(BaseSatelliteFetching):
     site = models.CharField(max_length=255)
+    model_run_uuid = models.CharField(max_length=255, blank=True, null=True)
```
mvandenburgh, line 8: Is there ever a case where this would be null? And if not - we haven't deployed any of this code yet, so I wonder if it would be better to instead tweak the existing migration to include this field and set it to non-null. What do you think?

django/src/rdwatch_scoring/views/model_run.py
```diff
@@ -169,12 +188,20 @@ def list_model_runs(
     request: HttpRequest,
     filters: ModelRunFilterSchema = Query(...),  # noqa: B008
 ):
-    return filters.filter(get_queryset())
+    data = filters.filter(get_queryset())
+    # lets get the downloading information for model_runs
```
mvandenburgh, line 192:
```suggestion
```

django/src/rdwatch_scoring/models/satellite_fetching.py
BryonLewis: Good call.  I did it because I didn't feel like clearing data out of my existing DB when testing. I can add this stuff to the previous migration and revert it back to remove the blank/null.

mvandenburgh (approved):

PR #313, opened 2023-11-08
Scoring Coloring from T&E Database

BryonLewis:
Notes:

- Added a `Case(` to the scoring vector_tile endpoint for the color_code parameter.  It needs to check two different tables based on if it is ground truth or not.
- Modified Vector Tile to also return more data for the `sites` so a popup can be used properly.  This included stuff like configuraiton_name, and performer_name
- I removed a lot of the older ScoreColoring logic from the front-end because it is no longer needed.
- Refactored the `drawObservations` and `drawSitePolygons` to be undefined or an array consisting of 'model or 'groundtruth'.  This allows for different combinations of the 4 states to be drawn at once.
- LayerSelection.vue component is used to manage the layers being shown.
- I added the performerMap to the store because I needed a reference to the performer name instead of the ID for some stuff being used in LayerSelection.
- I had to add a site-fille-layer to the rendering for popup windows to work properly with sites themsevles.
- I modified the Popup rendering to pass props from both the site observations and the sit models.  In the future I may modify it further to make it a bit more performant and work with mousemove inside of sites.

### DIFF
--- a/vue/src/components/LayerSelection.vue
+++ b/vue/src/components/LayerSelection.vue
@@ -127,7 +127,6 @@ const toggleScoring = (data? : undefined | 'simple' | 'detailed') => {
           class="px-2 mx-2"
           v-bind="props"
           size="large"
-          :variant="!state.filters.drawObservations ? 'outlined' : undefined"
           :color="state.filters.drawObservations ? 'rgb(37, 99, 235)' : ''"
           @click="toggleObs()"
         >
@@ -177,7 +176,6 @@ const toggleScoring = (data? : undefined | 'simple' | 'detailed') => {
           class="px-2 mx-2"
           v-bind="props"
           size="large"
-          :variant="!state.filters.drawSiteOutline ? 'outlined' : undefined"
           :color="state.filters.drawSiteOutline ? 'rgb(37, 99, 235)' : ''"
           @click="toggleSite()"
         >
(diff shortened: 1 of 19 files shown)


vue/src/mapstyle/annotationStyles.ts
```diff
@@ -101,26 +101,135 @@ const styles: AnnotationStyle = {
...
+  25: {
+    color: 'magenta',
+    hex: '#D53FFF',
+    title: 'False Positive (Proposal)',
+    description:'False Positive (Proposal)',
+    simple: false,
+    detailed: true,
+    id: 25,
+  },
+}
```
mvandenburgh, line 222: Would we want to add an endpoint for fetching these from the server instead of hardcoding them here? (In the future, not saying this should be done on this PR)

vue/src/components/LayerSelection.vue
```diff
@@ -0,0 +1,290 @@
...
+import { ApiService } from "../client";
+import { state } from "../store";
+
+
+
+const scoringApp = computed(()=> ApiService.apiPrefix.includes('scoring'));
+
+
+
+
```
mvandenburgh, line 12:
```suggestion
```

mvandenburgh (approved):
LGTM, pulled and tested and it works as expected.

mvandenburgh (approved):

PR #319, opened 2023-11-16
Fix date picker on download image prompt

Sean-S-Wang:
Fix the date picker closing the menu if anything except a day on the calendar was clicked


BryonLewis (approved):
Thanks for adding this.

PR #320, opened 2023-11-21
Add global error handler/popup snackbar

mvandenburgh:
This configures the Vue app with a global error handler that will catch any JavaScript errors that are thrown and render them in a UI popup. A copy-to-clipboard button is included to make including these errors in bug reports easier.

Here's an example of a background request failing:

https://github.com/ResonantGeoData/RD-WATCH/assets/37340715/d94be735-5dca-4ac4-a9cf-d4f01e7d5631

And here's an example of an error with a longer stack trace to demonstrate the scrolling functionality (i inserted a `throw Error('...')` manually here)

https://github.com/ResonantGeoData/RD-WATCH/assets/37340715/2dd6ef66-2c57-43ed-8132-46c198df8ca4

### DIFF
--- a/vue/src/components/ErrorPopup.vue
+++ b/vue/src/components/ErrorPopup.vue
@@ -1,0 +1,64 @@
+<script setup lang="ts">
+defineProps<{
+  errorText: string
+}>();
+
+const emit = defineEmits<{
+  (e: "close"): void;
+}>();
+
+function copyToClipboard(text: string) {
+  navigator.clipboard.writeText(text);
+}
+</script>
+
+<template>
+  <v-snackbar
+    absolute
+    multi-line
+    location="top right"
+    :model-value="!!errorText"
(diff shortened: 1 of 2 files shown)


vue/src/components/ErrorPopup.vue
```diff
@@ -0,0 +1,64 @@
...
+        class="overflow-auto"
+        style="max-height: 40vh; white-space: pre-line;"
+      >
+        {{ errorText }}
+      </div>
+    </div>
+    <template #actions>
+      <v-btn
+        color="info"
+        variant="text"
```
BryonLewis, line 57: very minor, either switching it to variant='flat' or change the color so that info color of Blue doesn't clash as much with the red background.

BryonLewis (approved):
one extremely minor thing regarding the close button in the component.

PR #321, opened 2023-11-21
Clean up new site endpoint

mvandenburgh:
I made a couple suggested changes to #318 here, organized by commit. @BryonLewis any thoughts on these changes? If you agree with them, I can push changes for the frontend as well.


BryonLewis:
I'm fine with it.  It's going to create some work in that fact that you can't rely anymore on `/scoring/site` and `/site` because of our messed up way of using site_evaluation under the /evaluations/ endpoint.  Also I think you may need to sync up if you want to have /details on the end of the endpoint or not, because that's different between the two as well.

BryonLewis (approved):
I'm okay with it if you want to change these endpoints.  Just my comments that it will make some front-end /scoring vs /non-scoring endpoints a little more difficult.  I think the real solution in the future is to move the stuff in /evaluations/ to /sites/ to be more in line with the scoring DB.

mvandenburgh:
I'm going to close this as it's become significantly outdated and isn't really relevant anymore.

PR #322, opened 2023-11-24
Optimize cancellation endpoint to avoid timeouts

mvandenburgh:
Iterating over every `SiteEvaluation` inside an endpoint can potentially spike response times. In extreme cases, this tends to cause gateway timeout errors on the AWS deployment if run on a model run with thousands (or more) of associated site evaluations. To fix this, I moved most of the cancellation logic into a celery task.


BryonLewis (approved):
Thanks

PR #326, opened 2023-12-04
Add debouncing to some buttons

mvandenburgh:
This prevents unnecessary API requests from being fired if the user clicks a button multiple times.


BryonLewis (approved):
I might even increase that delay to 5000 to give the server more time to respond to what is happening.

PR #328, opened 2023-12-04
fix null site obs labels

BryonLewis:
Allows for null labels in the SiteObservations.  There were specific sites which had this value that were failing before.

### DIFF
--- a/django/src/rdwatch_scoring/views/site_image.py
+++ b/django/src/rdwatch_scoring/views/site_image.py
@@ -7,7 +7,6 @@ from django.core.files.storage import default_storage
 from django.db.models import Count, F, Func, Value
 from django.db.models.functions import JSONObject
 from django.http import HttpRequest
-from django.shortcuts import get_object_or_404
 
 from rdwatch.db.functions import BoundingBox, ExtractEpoch
 from rdwatch.views.site_image import SiteImageResponse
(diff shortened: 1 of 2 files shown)


django/src/rdwatch_scoring/views/site_image.py
```diff
@@ -125,8 +102,6 @@ def site_images(request: HttpRequest, id: UUID4):
     output['label'] = site_eval_data['json']['label']
     output['status'] = site_eval_data['json']['status']
     output['notes'] = site_eval_data['json']['notes']
-    if ground_truth.exists():
```
mvandenburgh, line 128: So this was always returning False before?

mvandenburgh (approved):

django/src/rdwatch_scoring/views/site_image.py
BryonLewis: No, it was returning true because it was grabbing a list of the same Sites that matched the region and performer name and just returning the first one.

Our RGD endpoint worked in a similar fashion by matching the names between siteIds to determine the ground truth.  That doesn't work here, you really need to use the BAS Proposal table for the source ID, then look at the `site_truth_matched` which is an array of siteIds (in the KR_R001_KIT_V2.20 style)  then find the geometry for them. This would have required then re-factoring the front-end to draw multiple groundtruth sites on the image.  I felt it wasn't worth it and removed the whole thing.  Ground_truth can be null/None so not having it is better for right now.

PR #329, opened 2023-12-05
Redesign error popup UI

mvandenburgh:
Moves the error popup into a dialog that the user must open themselves. When an error occurs, the button in the top right corner will become clickable. If the user clicks it, a popup dialog will appear with the error text. 

Demo:

error1

### DIFF
--- a/vue/src/components/ErrorPopup.vue
+++ b/vue/src/components/ErrorPopup.vue
@@ -10,9 +10,10 @@ function copyToClipboard(text: string) {
   <v-dialog width="500">
     <template #activator="{ props }">
       <v-btn
+        v-show="state.errorText"
         v-bind="props"
         icon="mdi-alert"
-        location="top right"
+        location="top left"
         position="absolute"
         class="ma-4"
         :color="state.errorText ? 'warning' : 'default'"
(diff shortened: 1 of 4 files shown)


mvandenburgh:
I also added https://github.com/ResonantGeoData/RD-WATCH/pull/329/commits/3598a76d7e5273965f03724d736703e17a30d4fe, which makes it so the entire `Error` object is dumped to the user's clipboard. This, combined with enabling the Django `DEBUG` flag, will ensure we get useful error traces from the server.

vue/src/components/ErrorPopup.vue
```diff
@@ -13,52 +13,61 @@ function copyToClipboard(text: string) {
...
-        color="info"
-        variant="flat"
-        @click="emit('close')"
-      >
-        Close
-      </v-btn>
+        v-bind="props"
+        icon="mdi-alert"
+        location="top right"
+        position="absolute"
```
BryonLewis, line 22: Do we want this absolute and in the upper right so that even when the Site Details bar is open it will cover up this icon?
Should it maybe be placed in the upper left, now that there is room around the Resonant GeoData icon?
Then maybe it can have an invisible state unless there is something in the error log?

Just an idea:
image

BryonLewis (changes requested):

vue/src/components/ErrorPopup.vue
mvandenburgh: Oh I like that better too. I'll move it there

PR #331, opened 2023-12-05
Fix debounce delay

mvandenburgh:


mvandenburgh:
I tested this locally and confirmed it works as expected :+1:

PR #334, opened 2023-12-11
Fix memory usage in delete task

mvandenburgh:
I noticed model runs with an expiration time on them weren't being deleted as expected. Looking at the AWS container logs, something is causing a OOMKilled error to occur. I didn't know this was a thing, but apparently calling `.delete()` on a QuerySet will (most of the time at least) cause Django to load each object into memory. I expected it to just do a `DELETE` query, but I guess Django has to do some extra housekeeping with signals and cascades (see [here](https://stackoverflow.com/questions/31477723/how-to-prevent-django-from-loading-objects-in-memory-when-using-delete) for more info).

This PR refactors that task to do deletions in batches, to avoid loading every model run/site evalutation/site observation that's being deleted into memory at once.

### DIFF
--- a/django/poetry.lock
+++ b/django/poetry.lock
@@ -1165,6 +1165,17 @@ certifi = "*"
 urllib3 = "*"
 
 [[package]]
+name = "more-itertools"
+version = "10.1.0"
+description = "More routines for operating on iterables, beyond itertools"
+optional = false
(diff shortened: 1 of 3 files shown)


BryonLewis (approved):
:+1:

PR #337, opened 2023-12-12
Add mode filter

mvandenburgh:
Adds a new filter for evaluation run modes.

mode

Depends on #336 being merged first.


mvandenburgh:
@BryonLewis I updated this to only show the filter in scoring mode as we discussed

BryonLewis:
Did we want the ability to have both batch and incremental be selected?

mvandenburgh:
> Did we want the ability to have both batch and incremental be selected?

You're right, my bad. I'll update that

BryonLewis (approved):
LGTM

PR #338, opened 2023-12-12
DRAFT: Ordering/filtering of model runs and proposed sites fix

Sean-S-Wang:
Addresses these issues
https://smartgitlab.com/TE/RD-WATCH/-/issues/50 - Filter model runs by evaluation
https://smartgitlab.com/TE/RD-WATCH/-/issues/52 - Order model runs by evaluation, region, performer
https://smartgitlab.com/TE/RD-WATCH/-/issues/53 - Number of sites should be number of proposed sites

PR #340, opened 2023-12-12
1.3 eval filter

BryonLewis:
Taking #338 and made some changes for the UI and fixing some Warnings/errors

- No Eval for RGD endpoint error
- fixes Performer warnings for recursive lookups
- ran pre-commit and linting
- Added page sizes for region endpoints of 1000 (when loading all annotations in RGD you get more than 100 regions).

Changed some vuetify logic to make the filters look like this: 
image

### DIFF
--- a/django/src/rdwatch_scoring/api.py
+++ b/django/src/rdwatch_scoring/api.py
@@ -14,7 +14,6 @@ api.add_router('/sites/', views.site.router)
 api.add_router('/evals/', views.eval.router)
 
 
-
 # useful for getting information back about validation errors
 @api.exception_handler(ValidationError)
 def custom_validation_errors(request, exc):
--- a/django/src/rdwatch_scoring/api.py
+++ b/django/src/rdwatch_scoring/api.py
@@ -11,7 +11,7 @@ api.add_router('/observations/', views.observation.router)
 api.add_router('/performers/', views.performer.router)
 api.add_router('/regions/', views.region.router)
 api.add_router('/sites/', views.site.router)
-api.add_router('/evals/', views.eval.router)
+api.add_router('/eval-numbers/', views.eval.router)
 
 
 # useful for getting information back about validation errors
--- a/django/src/rdwatch_scoring/api.py
+++ b/django/src/rdwatch_scoring/api.py
@@ -11,6 +11,8 @@ api.add_router('/observations/', views.observation.router)
 api.add_router('/performers/', views.performer.router)
 api.add_router('/regions/', views.region.router)
 api.add_router('/sites/', views.site.router)
+api.add_router('/evals/', views.eval.router)
+
 
 
 # useful for getting information back about validation errors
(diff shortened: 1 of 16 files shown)


django/src/rdwatch_scoring/views/eval.py
```diff
@@ -0,0 +1,26 @@
+from ninja.pagination import RouterPaginated
+
+from django.db.models import CharField, Value
+from django.db.models.functions import Concat
+from django.http import HttpRequest
+
+from rdwatch_scoring.models import EvaluationRun
+
+router = RouterPaginated()
```
BryonLewis, line 9: Should this be updated for the case where there are more than 100 evaluations?  I did some stuff to default regions to 1000, should this be increased from the stock 100?

BryonLewis (commented):
Couple of internal comments on pagination and if we should change the name `/eval/` to something like `/evalnum/` to it isn't confused with evaluationRuns.

django/src/rdwatch_scoring/views/eval.py
```diff
@@ -0,0 +1,26 @@
...
+    return (
+        EvaluationRun.objects.annotate(
+            eval=Concat(
+                'evaluation_number',
+                Value('.'),
+                'evaluation_run_number',
+                output_field=CharField(),
+            )
+        )
+        .order_by('evaluation_number', 'evaluation_run_number')
```
BryonLewis, line 23: I think we may actually want these in descending order.  So the newest is at the top?

django/src/rdwatch_scoring/api.py
```diff
@@ -11,6 +11,7 @@
 api.add_router('/performers/', views.performer.router)
 api.add_router('/regions/', views.region.router)
 api.add_router('/sites/', views.site.router)
+api.add_router('/evals/', views.eval.router)
```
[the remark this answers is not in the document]
mvandenburgh: I agree, `/evalnums` or `/eval-nums` would be clearer

django/src/rdwatch_scoring/views/eval.py
mvandenburgh: I think that would be fine. 100 vs 1000 wouldn't make a huge difference i would think

django/src/rdwatch_scoring/views/eval.py
mvandenburgh: Agreed.

django/src/rdwatch_scoring/views/eval.py
mvandenburgh: Oh, i see this is already done

mvandenburgh:
Updated this to 
1) Change `/evals` endpoint to `/eval-numbers`	for clarity
2) Order evaluation numbers in descending order, so the newest one comes first.

mvandenburgh (approved):

PR #342, opened 2023-12-18
Docker Client Dev Service

BryonLewis:
- Adds a rdwatch-client service to the docker-compose.override.yml
- I need to mount the full app because we need the git information for the client to display the hash.
- I copy over the folder (mounted as read only) so npm can install node_modules without messing with the user's local permissions
- There is a git config setting to allow it to read data that isn't owned by the user for the client hash displaying
- NPM install is ran and then npm run dev to launch the server
- vite config is modified with an environment variable `VITE_DOCKER_DEVELOPMENT` which tells it to proxy to the `django` container instead of `localhost`.
- If the `VITE_DOCKER_DEVELOPMENT` is false it will instead run the dev server at `localhost:3000` so it isn't conflicting with the docker compose service port.

I'm all for any argument you have about making another Dockerfile and using that to build the container instead of doing it at launch each time through the commands like I am.   I just figured that this would mainly be used for demoing and development so having an up-to-date version of the client code would always make sense.  We could always build another image and make it part of our Github container registry.

### DIFF
--- a/vue/vite.config.ts
+++ b/vue/vite.config.ts
@@ -16,14 +16,14 @@ export default ({ mode}) => {
     .toString()
     .trimEnd();
 
-    process.env.VITE_GIT_COMMIT_DATE = commitDate;
-    process.env.VITE_GIT_BRANCH_NAME = branchName;
-    process.env.VITE_GIT_COMMIT_HASH = commitHash;
-    process.env.VITE_GIT_LAST_COMMIT_MESSAGE = lastCommitMessage;
-    const dockerDev = process.env.VITE_DOCKER_DEVELOPMENT;
-    // Change host to django when running vite inside docker
-    const devHost = dockerDev ? 'django' : 'localhost';
-    const devPort = dockerDev ? 8080 : 3000;
+  process.env.VITE_GIT_COMMIT_DATE = commitDate;
+  process.env.VITE_GIT_BRANCH_NAME = branchName;
+  process.env.VITE_GIT_COMMIT_HASH = commitHash;
+  process.env.VITE_GIT_LAST_COMMIT_MESSAGE = lastCommitMessage;
+  const dockerDev = process.env.VITE_DOCKER_DEVELOPMENT;
+  // Change host to django when running vite inside docker
+  const devHost = dockerDev ? 'django' : 'localhost';
+  const devPort = dockerDev ? 8080 : 3000;
   return defineConfig({
       plugins: [vue()],
       server: {
--- a/vue/vite.config.ts
+++ b/vue/vite.config.ts
@@ -1,35 +1,41 @@
-import { defineConfig } from "vite";
+import { defineConfig, loadEnv } from "vite";
 import vue from "@vitejs/plugin-vue";
 import { execSync } from "child_process";
 
-export default defineConfig(() => {
-  // Load env file based on `mode` in the current working directory.
-  // Set the third parameter to '' to load all env regardless of the `VITE_` prefix.
-  const commitDate = execSync("git log -1 --format=%cI").toString().trimEnd();
-  const branchName = execSync("git rev-parse --abbrev-ref HEAD")
-    .toString()
-    .trimEnd();
-  const commitHash = execSync("git rev-parse HEAD").toString().trimEnd();
-  const lastCommitMessage = execSync("git show -s --format=%s")
-    .toString()
-    .trimEnd();
+export default ({ mode}) => {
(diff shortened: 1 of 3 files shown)


vue/vite.config.ts
```diff
@@ -1,35 +1,41 @@
...
+    // Load env file based on `mode` in the current working directory.
+    // Set the third parameter to '' to load all env regardless of the `VITE_` prefix.
+    const commitDate = execSync("git log -1 --format=%cI").toString().trimEnd();
+    const branchName = execSync("git rev-parse --abbrev-ref HEAD")
+      .toString()
+      .trimEnd();
+    const commitHash = execSync("git rev-parse HEAD").toString().trimEnd();
+    const lastCommitMessage = execSync("git show -s --format=%s")
+      .toString()
+      .trimEnd();
```
mvandenburgh, line 17: Just fixing some inconsistent whitespace -
```suggestion
  // Load env file based on `mode` in the current working directory.
  // Set the third parameter to '' to load all env regardless of the `VITE_` prefix.
  const commitDate = execSync("git log -1 --format=%cI").toString().trimEnd();
  const branchName = execSync("git rev-parse --abbrev-ref HEAD")
    .toString()
    .trimEnd();
  const commitHash = execSync("git rev-parse HEAD").toString().trimEnd();
  const lastCommitMessage = execSync("git show -s --format=%s")
    .toString()
    .trimEnd();
```

vue/vite.config.ts
```diff
@@ -1,35 +1,41 @@
...
-          target: "http://localhost:8000",
-          xfwd: true,
+    process.env.VITE_GIT_COMMIT_DATE = commitDate;
+    process.env.VITE_GIT_BRANCH_NAME = branchName;
+    process.env.VITE_GIT_COMMIT_HASH = commitHash;
+    process.env.VITE_GIT_LAST_COMMIT_MESSAGE = lastCommitMessage;
+    const dockerDev = process.env.VITE_DOCKER_DEVELOPMENT;
+    // Change host to django when running vite inside docker
+    const devHost = dockerDev ? 'django' : 'localhost';
+    const devPort = dockerDev ? 8080 : 3000;
```
mvandenburgh, line 26: Just fixing some inconsistent whitespace -
```suggestion
  process.env.VITE_GIT_COMMIT_DATE = commitDate;
  process.env.VITE_GIT_BRANCH_NAME = branchName;
  process.env.VITE_GIT_COMMIT_HASH = commitHash;
  process.env.VITE_GIT_LAST_COMMIT_MESSAGE = lastCommitMessage;
  const dockerDev = process.env.VITE_DOCKER_DEVELOPMENT;
  // Change host to django when running vite inside docker
  const devHost = dockerDev ? 'django' : 'localhost';
  const devPort = dockerDev ? 8080 : 3000;
```

vue/vite.config.ts
```diff
@@ -1,35 +1,41 @@
...
-      port: 8080,
-      proxy: {
-        "/api": {
-          target: "http://localhost:8000",
-          xfwd: true,
+    process.env.VITE_GIT_COMMIT_DATE = commitDate;
+    process.env.VITE_GIT_BRANCH_NAME = branchName;
+    process.env.VITE_GIT_COMMIT_HASH = commitHash;
+    process.env.VITE_GIT_LAST_COMMIT_MESSAGE = lastCommitMessage;
+    const dockerDev = process.env.VITE_DOCKER_DEVELOPMENT;
```
mvandenburgh, line 23: What do you think about making this env var be something like `VITE_DEV_PROXY_TARGET`, and make it default to `localhost:8000`, and override it to `django:8000` in the docker-compose client service? Then, we don't need to add any conditional logic to the vite.config.ts which seems cleaner to me.

BryonLewis: I think we still need conditionals (in the simplified form of `process.env.VITE_DEV_PROXY_TARGET || 'localhost'` ) unless we add a separate .env to the client directory for the default localhost when you decide to run `npm run dev` outside of the container.

Or I may be misinterpreting what you mean here?


I also want it to explicitly use 8080 for the dev port in the container and 3000 when running outside the container.  Mostly to prevent port conflicts if you happen to be changing node package dependencies live and don't want to relaunch the container for it to update.

mvandenburgh: Ah, I forgot you were trying to get it to default to 3000 when outside the container. In that case, I think my approach isn't any simpler. So i think the way you had it before is fine, or we can just leave it as is now, I'm fine with either.

mvandenburgh (approved):

PR #343, opened 2023-12-19
Image Viewer Rescaling 

BryonLewis:
When using the rescaling function on the image-viewer there were complaints that it was difficult to see the S2/L8 images because they were scaled up to match the geospatial bounds of the WV images.

This PR adds an option in the settings to zoom out the rescaled images to provide more context for the S2/L8 images (up to 5X the size of the box).

Since we don't have any additional data for the WorldView image outside of it's bounds it is marked as gray outside of the area.

Note: 
- Zooming out on a mix of WV and S2 can result in a large GIF because I'm still displaying the WV at native resolution.  I.E if the WorldView resolution is 500x500 and you zoom out 2X then all images in the GIF sequence are now 1000x1000 with the S2/L8 images upscaled to fit there.


BryonLewis:
I get that logic can seem a bit complex between scaling.  I would just suggest taking a site downloading  WV/S2 images and then playing with the rescale toggle setting
image

And then under the Cog Wheel adjust the zoom out setting.

mvandenburgh (approved):

PR #347, opened 2024-01-08
Proposal Fixes and Improvments

### LINKED ISSUE #346: Proposal View Fixes
- [x] Ground Truth range disabling if not matched.  Currently with null/null time ranges the ground truth is displayed for sites when it shouldn't be.
- [x] Add toggling for view of the models with the selected region.  Currently, it doesn't have the options for the sites/observations or scoring fields.  This needs to be added to the proposal view.
- [x] A way to toggle on the ground truth for all sites for a specific region associated with the proposal view.  
- [x] GroundTruth vs proposal sites should have some shading/opacity to differentiate between them.
- [x] You should be able to click on a proposal site to bring up the detailed view for them.
- [x] Clicking in an empty area should dismiss the container and return back to the full view?
- [x] Hovering over a site in the map or hovering over the site in the list should somehow indicate the relationship.
- [x] Accepted/Rejected sites should have some indication on the map as well like an opacity selection
- [x] Add a toggle to the Accepted/Rejected visibility on the map.
- [x] Add a toggle to turn on/off shading for groundtruth in the proposal view
- [x] Recenter the Date filter for the Proposal view
- [x] Add in Satellite region view for proposals.
- [x] Add in buttons for downloading images individually for a site.
- [x] Probably need to include some status for downloading when getting proposals and a way to view it
- [x] fix adjudicated listing in model-run being tied to the caching system, it may need to be moved into a separate area
- [x] Add scrolling to hovered over proposal site.
- [x] Test ground truth matching for proposals.
- [x] Updated vector tile caching to use a modified_timestamp so it will update when people change the siteEvaluation data when in proposal mode.  Before if you change the type of a site it wouldn't update in the map view.
- [x] fix client side caching of vector tiles when changing data.

BryonLewis:
resolves #346 

list of changes below:
- Ground Truth range disabling if not matched.  Currently with null/null time ranges the ground truth is displayed for sites when it shouldn't be.
- Add toggling for view of the models with the selected region.  Currently, it doesn't have the options for the sites/observations or scoring fields.  This needs to be added to the proposal view.
- A way to toggle on the ground truth for all sites for a specific region associated with the proposal view.  
- GroundTruth vs proposal sites should have some shading/opacity to differentiate between them.
- You should be able to click on a proposal site to bring up the detailed view for them.
- Clicking in an empty area should dismiss the container and return back to the full view?
- Hovering over a site in the map or hovering over the site in the list should somehow indicate the relationship.
- Accepted/Rejected sites should have some indication on the map as well like an opacity selection
- Add a toggle to the Accepted/Rejected visibility on the map.
- Add a toggle to turn on/off shading for groundtruth in the proposal view
- Recenter the Date filter for the Proposal view
- Add in Satellite region view for proposals.
- Add in buttons for downloading images individually for a site.
- Probably need to include some status for downloading when getting proposals and a way to view it
- fix adjudicated listing in model-run being tied to the caching system, it may need to be moved into a separate area
- Add scrolling to hovered over proposal site.
- Test ground truth matching for proposals.
- Updated vector tile caching to use a modified_timestamp so it will update when people change the siteEvaluation data when in proposal mode.  Before if you change the type of a site it wouldn't update in the map view.

Notes:

- I turned off model-run caching for proposal endpoints.  This is due to the adjudicated number to indicate how many are complete.  We could additionally move this out of the queryset and into a separate pass for proposals if it would speed things up while keeping the generic cache
- I added a 'groundTruth' field to the adjudicated fields for a model run.  This is used to help indicate which file is the ground truth associated with a proposal
- Vector-tile caching has been changed to utilize a 'modified' date instead of the creation date.  This also includes adding a modified timestamp to the evaluation model.  This was done because it helps change the vector tile after it's been modified by the user.

TODO
- [x] fix client side caching of vector tiles when changing data.

https://github.com/ResonantGeoData/RD-WATCH/assets/61746913/be13623a-fa6f-49a4-8f10-415f978b7ceb

Small 20240119 Update:

- Per one of the T&E issues I swapped the export from `.json` to `.geojson`.


django/src/rdwatch/views/model_run.py
```diff
@@ -229,6 +231,13 @@ def get_queryset():
                     then=JSONObject(
                         proposed=Coalesce(Subquery(proposed_count_subquery), 0),
                         other=Coalesce(Subquery(other_count_subquery), 0),
+                        groundTruths=Subquery(
```
mvandenburgh, line 234:
```suggestion
                        ground_truths=Subquery(
```

For consistency with the case of other fields in this query.

django/src/rdwatch/views/model_run.py
```diff
@@ -425,19 +438,26 @@ def get_region(model_run_id: UUID4):
...
+            downloading=Exists(
+                SatelliteFetching.objects.filter(
+                    site=OuterRef('pk'),
+                    status=SatelliteFetching.Status.RUNNING,
+                )
+            ),
         )
         .aggregate(
-            evaluations=JSONBAgg(
+            proposedSites=JSONBAgg(
```
mvandenburgh, line 460:
```suggestion
            proposed_sites=JSONBAgg(
```

django/src/rdwatch/models/site_evaluation.py
```diff
@@ -19,6 +19,10 @@ class SiteEvaluation(models.Model):
     id = models.UUIDField(primary_key=True, default=uuid4, editable=False)
 
     timestamp = CreationDateTimeField()
+    modified_timestamp = models.DateTimeField(
+        auto_now=True,
+        help_text='Timestamp of the last modification',
+    )
```
mvandenburgh, line 25: Using a modified timestamp makes sense logically, but unfortunately there is a major issue with the `auto_now` option. Django sort of implies that this is a magic column that will always get updated when the row is modified, but sadly that's not the case - `modified_timestamp` is updated when calling .save() on a model, but it will not be updated if .update() is used. See https://docs.djangoproject.com/en/5.0/ref/models/fields/#django.db.models.DateField.auto_now

When we tried using this feature on other projects this ended up causing some weird data inconsistency bugs as the codebase got more and more complex. In the end we ended up having rewriting things to remove it and explicitly update the `modified_timestamp` field when needed to remove any ambiguity. I don't think we're necessarily at that point yet (and maybe we'll never get there, since at the end of the day these are only used for caching purposes, it's not like they're user facing or there's any other functionality depending on them), but I think we should consider using the explicit update approach instead - any thoughts?

mvandenburgh (commented):
Some initial review comments just from reading through the code. I'll clone this locally to test it and make a follow-on review for that.

django/src/rdwatch/models/site_evaluation.py
BryonLewis: I am all for being explicit if there may be some confusion.  I've left it as  standard DateTime field, added a comment about manually modifying it.  Then added the updates to the creationTime and the patch endpoints for siteEvaluations.

vue/src/client/services/ApiService.ts
```diff
@@ -482,6 +483,7 @@ export class ApiService {
   }
 
   public static patchSiteEvaluation(id: string, data: SiteEvaluationUpdateQuery): CancelablePromise<boolean> {
+    // We clear the cache so it will reload the tiles and display the new color.
```
mvandenburgh, line 486: Is this comment misplaced?

vue/src/components/LayerSelection.vue
```diff
@@ -1,18 +1,55 @@
...
+          state.openedModelRuns.add(id);
+        }
+        const configuration_id = Array.from(state.openedModelRuns);
+        state.filters = { ...state.filters, configuration_id};
+      }
+      else if (!state.filters.drawObservations?.includes('groundtruth') && !state.filters.drawSiteOutline?.includes('groundtruth')) {
+        if (state.openedModelRuns) {
+          if (state.openedModelRuns.has(id)) {
+            state.openedModelRuns.delete(id);
+          }
```
mvandenburgh, line 26:
```suggestion
          state.openedModelRuns.delete(id);
```

Calling `Set.has()` before `Set.delete()` isn't necessary - [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/delete)

mvandenburgh (approved):
Thanks for the video demo, that was really helpful to figure out what the new features are. I noticed a couple other minor code issues that I flagged in the review above, otherwise LGTM

vue/src/client/services/ApiService.ts
BryonLewis: I removed that I think the `state.filters.randomKey` plus the comment above it indicates the clearing of the tile cache.
I may have meant that on the server it clears the vector tile cache when items are patched, but I think that's clear in the endpoint in django now.

PR #348, opened 2024-01-17
[draft] Microservice for Vector Tile server

mvandenburgh:
Some notes:
- I thought it made more sense for the tileserver application code to live in the RD-WATCH repo, but that presents an issue in that the definition of the Lambda Function in the private infrastructure repo needs to somehow reference the code in this repo. I solved this by refactoring the tile serving code into an npm module, and added a workflow to push it to github packages' NPM registry. The Lambda Function in the infrastructure repo will then have this module as a dependency.


mvandenburgh:
Superseded by #357

PR #349, opened 2024-01-24
Polygon Editor

BryonLewis:
Starting the work  on the polygon editor:

Current Demo:

https://github.com/ResonantGeoData/RD-WATCH/assets/61746913/53e4d1a8-f13e-42fb-bbfe-7600abf17731

TODO:

- [x] Find a way to disable or hide the polygon editor control.  It seems as though a control needs to be added for basic polygon support but I think it would be better to hide it.
- [x] Create a global state for 'editingPolygon' and use that to disable the popup and other selection/deselection methods for proposals.
- [x] When a polygon is kicked into edit mode the vector-tile for that editing mode polygon needs to be disabled.  This could be done if the editingPolygons had some information stored in it.  Like the Site Id and use that to disable the vector tile rendering while it is in editing mode.  Probably need to add to the filters a editingPolygon that will disable the currently editable polygon
- [x] If a user jumps out of a current polygon while in editing mode it needs to automatically disable and revert back to the standard polygon.
- [x] The editing mode needs to have some tools for changing the styling, I used an example from the github for mapbox-gl-draw but it resulted in not rendering any points or the polygon.
- [x] On completion of the edit and clicking save it will turn the site back on and will bump the randomKey to refresh the cache for the visualization of the 
- [x] I think I want a toggle to display the site Polygon on the image instead of the observation polygon.  I think this would help a bit for when the polygon changes.
- [x] Figure out how to integrate point deletion into the system.  Thinking of a UI element that appears in the map which allows for deletion of a point.  I can add it to the userEditPolygon composition function and when a point is selected the user can edit it.


BryonLewis:
when merged with #355 the CI should complete properly.

mvandenburgh (approved):
LGTM, tested and it works as expected :+1:

PR #350, opened 2024-01-24
Notes Update

BryonLewis:
Per an issue in gitlab about notes overflowing and not having enough room to type notes.

- switched to truncate: ellipsis for the notes with a tooltip hover to show the full note
- The v-text-field was change to a v-textarea to allow easier editing.

image

image


mvandenburgh (approved):
LGTM. Python linting failures seeme unrelated to this PR, i'll look into those..

PR #356, opened 2024-02-07
Sam Integration

BryonLewis:
Starting of the SAM integation testing

20240213:

- ~~Check why the pre-signed URL npy file isn't loading properly.  Test it in the VueSAM application to see if the file is proper or if it is something with the way that the S3 presigned URL is serving the data instead that is causing the problem.~~
- Take a look and see if the VueSAM demo code can be better integrated into the system on the map for hovering and selecting segments
- ~~Currently there is no status/in-progress for generating a npy/image embedding file.  I may want to update that in the future to use some of the satellite fetching to use that.~~
- Segement Selection needs to be additive so it will add to the existing modules.  Also need a way to remove selections like right click.
- Add code in to convert image based polygon back into geospatial polygon.

20240219:

- [x] Swap the regular Canvas for a SAM canvas in the SAM editing mode.
- [x] Fix scaling issues with this new SAM canvas mode of a single image
- [x] Add capability to add/subtract segments
- [x] Completion button is clicked and transfers the current polygon to the editing field.

20240304:

- [x] Refactor the tools used for the editing of the segment validation.  I.E the layout doesn't look that great currently.  We have options for cancellation, generating the polygon undo and the clearing.  I need to space organize the buttons better.
- [x] I think it makes sense to only allow SAM on WV Images and disable it for S2 images.
- [x] I also believe now when using it that the filter settings for S2/WV/PL/L8 should be a global setting rather than being setting per image.  In conjunction with this I think the filter should show some indication of the limited values in the filter.


django/src/rdwatch/tasks/__init__.py
```diff
@@ -587,3 +591,53 @@ def generate_site_images_for_evaluation_run(
             scale,
             bboxScale,
         )
+
+
+@shared_task
+def generate_image_embedding(id: int):
+    site_image = SiteImage.objects.get(pk=id)
+    if site_image:
```
mvandenburgh, line 599: These two statements used together are redundant. Django will raise an exception if `SiteImage.objects.get(pk=id)` doesn't return a SiteImage; I _think_ there's no way for this task to be called if the `site_image` doesn't yet exist, correct? If that's true, I would say remove the conditional (which also has the advantage of decreasing the amount of indents in the code below)

django/src/rdwatch/tasks/__init__.py
```diff
@@ -587,3 +591,53 @@ def generate_site_images_for_evaluation_run(
...
+        with transaction.atomic():
```
mvandenburgh, line 600: What is the reason for using a transaction here?

django/src/rdwatch/tasks/__init__.py
```diff
@@ -587,3 +591,53 @@ def generate_site_images_for_evaluation_run(
...
+                    site_image.image_embedding.save(
+                        os.path.basename('sampleImage.npy'), ContentFile(embedding_data)
+                    )
+
+                    # Step 3: Save the SiteImage instance
+                    site_image.save()
+
+                # Step 4: Clean up local files
+                os.remove(local_file_path)
+                os.remove('sampleImage.npy')
```
mvandenburgh, line 639: Is there a reason to not use `tempfile.NamedTemporary` file here? I believe that would give us safer guarentees regarding placement and cleanup of these files, rather than doing it manually.

django/src/rdwatch/views/site_image.py
```diff
@@ -146,3 +151,71 @@ def site_images(request: HttpRequest, id: UUID4):
     output['evaluationGeoJSON'] = site_eval_data['json']['evaluationGeoJSON']
     output['evaluationBBox'] = site_eval_data['json']['evaluationBBox']
     return output
+
+
+@router.post('/{id}/image_embedding', response=UUID4)
+def post_image_embedding(request: HttpRequest, id: int):
+    if not SiteImage.objects.filter(pk=id).exists():
+        raise Http404()
```
mvandenburgh, line 159: This can be done more concisely with 
```suggestion
	get_object_or_404(SiteImage, pk=id)
```

django/src/rdwatch/views/site_image.py
```diff
@@ -146,3 +151,71 @@ def site_images(request: HttpRequest, id: UUID4):
...
+    else:
+        return generate_image_embedding.delay(id).id
+
+
+@router.get('/{id}/image_embedding_status/{uuid}')
+def get_image_embedding_status(request: HttpRequest, id: int, uuid: UUID4):
+    if not SiteImage.objects.filter(pk=id).exists():
+        raise Http404()
```
mvandenburgh, line 167: Same as above -
```suggestion
    get_object_or_404(SiteImage, pk=id)
```

django/src/rdwatch/views/site_image.py
```diff
@@ -146,3 +151,71 @@ def site_images(request: HttpRequest, id: UUID4):
...
+                'aws_location',
+                'image_embedding',
+                'pk',  # Assuming 'pk' is the primary key field name
+            )
+            .first()
+        )
+    except SiteImage.DoesNotExist:
+        raise Http404()
+
+    if site_image:
```
mvandenburgh, line 204: Isn't this check redundant, as we've already raised a `Http404` by this point if the `site_image` doesn't exist?

django/src/rdwatch/views/site_image.py
```diff
@@ -146,3 +151,71 @@ def site_images(request: HttpRequest, id: UUID4):
...
+                'percent_black',
+                'image_dimensions',
+                'aws_location',
+                'image_embedding',
+                'pk',  # Assuming 'pk' is the primary key field name
+            )
+            .first()
+        )
+    except SiteImage.DoesNotExist:
+        raise Http404()
```
mvandenburgh, line 202: I haven't tried this locally yet, but I _think_ you need to call `site_image.exists()` instead to do this check. `.filter` won't raise an exception if the query returns no result.

django/src/rdwatch/views/site_image.py
```diff
@@ -146,3 +151,71 @@ def site_images(request: HttpRequest, id: UUID4):
...
+            'cloudcover': site_image['cloudcover'],
+            'image': default_storage.url(site_image['image']),
+            'observation_id': site_image['observation_id'],
+            'percent_black': site_image['percent_black'],
+            'bbox': site_image['bbox'],
+            'image_dimensions': site_image['image_dimensions'],
+            'aws_location': site_image['aws_location'],
+            'image_embedding': default_storage.url(site_image['image_embedding']),
+            'id': site_image['pk'],
+        }
```
mvandenburgh, line 217: I think we should consider aliasing these in the `SiteImageSchema`, that there's no need to repeat all the values here and you can just do `return site_image` directly.

mvandenburgh (changes requested):
My plan is to do a couple review passes on this to break things up -
1) Review backend code (this review)
2) Review frontend code/vite changes
3) Pull the branch locally and test, and do another review if I encounter any issues or get any other thoughts from doing that

django/src/rdwatch/tasks/__init__.py
BryonLewis: Nope no reason, just was kind of in experimental mode and never converted it back after I got it working.  I'll change this.

django/src/rdwatch/views/site_image.py
BryonLewis: Just did some testing on it.  calling the 'first' on the filter will trigger the DoesNotExists if it returns an empty item.

django/src/rdwatch/views/site_image.py
BryonLewis: We can do most of this, except I think because we are using an older ninja version we might not be able to use resolvers for converting the image/image_embedding base URLS into resigned URLs.  Or there may be a way to do it and I just don't see it in the documentation.

django/src/rdwatch/views/site_image.py
```diff
@@ -148,3 +154,63 @@ def site_images(request: HttpRequest, id: UUID4):
     output['evaluationGeoJSON'] = site_eval_data['json']['evaluationGeoJSON']
     output['evaluationBBox'] = site_eval_data['json']['evaluationBBox']
     return output
+
+
+@router.post('/{id}/image_embedding', response=UUID4)
```
mvandenburgh, line 159:
```suggestion
@router.post('/{id}/image_embedding/', response=UUID4)
```

django/src/rdwatch/views/site_image.py
```diff
@@ -148,3 +154,63 @@ def site_images(request: HttpRequest, id: UUID4):
...
+def post_image_embedding(request: HttpRequest, id: int):
+    get_object_or_404(SiteImage, pk=id)
+    return generate_image_embedding.delay(id).id
+
+
+@router.get('/{id}/image_embedding_status/{uuid}')
```
mvandenburgh, line 165:
```suggestion
@router.get('/{id}/image_embedding_status/{uuid}/')
```

django/src/rdwatch/views/site_image.py
```diff
@@ -148,3 +154,63 @@ def site_images(request: HttpRequest, id: UUID4):
...
+    get_object_or_404(SiteImage, pk=id)
+    task = AsyncResult(uuid)
+    result = {
+        'state': task.state,
+        'status': task.status,
+    }
+    return result
+
+
+@router.get('/{id}/image', response=SiteImageSchema)
```
mvandenburgh, line 176:
```suggestion
@router.get('/{id}/image/', response=SiteImageSchema)
```

vue/vite.config.ts
```diff
@@ -21,20 +23,43 @@ export default ({ mode}) => {
...
+              dest: ".",
+            },
+          ],
+        }),
+        {
+          name: "configure-response-headers",
+          configureServer: (server) => {
+            server.middlewares.use((req, res, next) => {
+              // res.setHeader("Cross-Origin-Opener-Policy", "same-origin");
+              // res.setHeader("Cross-Origin-Embedder-Policy", "require-corp");
```
mvandenburgh, line 44:
```suggestion
```

django/src/rdwatch/views/site_image.py
mvandenburgh: Yeah, I'm not sure either. I'm fine waiting until we upgrade ninja to address this

mvandenburgh (commented):
Just some last nitpicks around the endpoint trailing slashes and removing comments.

BryonLewis:
In case you were wondering about the additional commits regarding the pyproject.toml and the lock file:

In order to reduce install size I limited poetry to only installing the CPU version of pytorch but that then had issues with tox and setuptools.  So I reverted it back.  Not sure why the github CI container won't build.  I haven't died too deep into the logs yet for that, but it builds locally across machines fine.

mvandenburgh:
> In order to reduce install size I limited poetry to only installing the CPU version of pytorch but that then had issues with tox and setuptools. So I reverted it back. Not sure why the github CI container won't build. I haven't died too deep into the logs yet for that, but it builds locally across machines fine.

The github actions UI is bugging out or something for me and saying the logs were deleted. But my suspicion was that we are exceeding the disk space on the runner. I haven't checked the size of the `.pth` file, but I tested this theory by commenting it out on a new branch (https://github.com/ResonantGeoData/RD-WATCH/commit/61fd374a978320ee90c58f3b5a00abb58a2d3e0e), and it passes (https://github.com/ResonantGeoData/RD-WATCH/actions/runs/8377546499/job/22939892134)

mvandenburgh (approved):

PR #359, opened 2024-02-17
Updated readme by adding introductory text, features, links to documentation etc. 

aashish24:


aashish24:
@BryonLewis @mvandenburgh take a look at the update README and let me know if you would like to make any changes. Also, I wonder if we should simply call it RDWATCH (consistent with GeoWATCH) as opposed to RD-WATCH. Thoughts?

BryonLewis (approved):

mvandenburgh (approved):

PR #361, opened 2024-02-21
Fix scoring DB query error

mvandenburgh:
It seems it's possible for the scoring DB to have the same color code repeated in the DB, which was causing a subquery error message in the vector tile generation.


mvandenburgh:
Superseded by #362

PR #364, opened 2024-02-28
Docker Vector-Tiler Server Updates

BryonLewis:
- Adds vector-server for docker compose development to locally have node.js based tileserver running.
- Updates the `randomKey` parameter to work properly with the editing of polygons in the proposal view.


BryonLewis:
You may need to update the infrastructure code to support the new `VITE_DEV_VECTOR_PROXY_TARGET` which is used to point the client to the proper proxiess in the dev server version.

mvandenburgh (approved):

PR #366, opened 2024-03-08
Run tileserver with hot reloading in docker

mvandenburgh:


BryonLewis (approved):
Yeah, I probably should have done this from the beginning.  When working yesterday I just relied on portainer and restarting the vectorserver container.

PR #368, opened 2024-03-12
Dependency updates

mvandenburgh:
I skimmed this page https://docs.djangoproject.com/en/5.0/releases/5.0/#backwards-incompatible-changes-in-5-0, and the only breaking changes I found were https://github.com/ResonantGeoData/RD-WATCH/pull/368/commits/7062c4991142d8ebec92d89176b6a7f00414bc93 and https://github.com/ResonantGeoData/RD-WATCH/pull/368/commits/ee555d9dfa352054a11039e5a8ce8118361e5d1b.

### DIFF
--- a/django/pyproject.toml
+++ b/django/pyproject.toml
@@ -26,7 +26,7 @@ packages = [
 
 [tool.poetry.dependencies]
 python = "~3.11.0"
-django = ">=4.1,<4.2"
+django = ">=5.0.3,<5.1"
 psycopg2 = "~2.9.3"
 redis = { version = "~4.3.4", extras = [ "hiredis" ] }
 uritemplate = "^4.1.1"
--- a/django/pyproject.toml
+++ b/django/pyproject.toml
@@ -40,7 +40,7 @@ pillow = "^9.5.0"
 django-configurations = {version = "^2.5", extras = ["database", "email"]}
 django-storages = {version = "^1.14.2", extras = ["s3"]}
 django-celery-results = "^2.5.1"
-pystac-client = "^0.7.5"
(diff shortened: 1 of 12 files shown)


django/pyproject.toml
```diff
@@ -25,23 +25,23 @@ packages = [
...
-rio-tiler = "5.0.3"
+rio-tiler = "5.0.3" # TODO: upgrade blocked on pydantic 2
 mercantile = "^1.2.1"
-django-ninja = "^0.21.0"
-celery = "^5.2.7"
-django-extensions = "^3.2.1"
+django-ninja = "~0.22.2" # TODO: upgrade blocked on pydantic 2
+celery = "^5.3.6"
+django-extensions = "^3.2.3"
 pillow = "^9.5.0"
```
mvandenburgh, line 39: Note - we also need to upgrade Pillow, but that one actually seems to have major API changes. So, I'll make a follow up PR that does that one by itself.

BryonLewis:
The celery beat container is not able to import the module `pytz`.

mvandenburgh:
> The celery beat container is not able to import the module `pytz`.

I just saw this too, and resolved by removing the state `celerybeat-schedule` file from my celery beat container. Like this -

```bash
$ docker compose run --rm celery_beat bash
root@92d0c9d107dd:/app/django# rm celerybeat-schedule*
```

A complete rebuild of the container without caching should also work.

mvandenburgh:
(For some additional background - that seems to be a consequence of `celery` being upgraded)

BryonLewis (approved):
:+1:

PR #370, opened 2024-03-13
Include current version in server status endpoint, swagger page

mvandenburgh:
It was starting to become difficult to keep track of what's deployed on the different deployments we have. Now that we're referring to each release by it's version number, I think it's much better to use that for versioning instead of the full commit hash.

I didn't put this in the UI yet, but maybe we should rework that at some point to hit the `/api/status` endpoint instead of statically bundling that info in the JS code. Maybe something to think about in the future when we have more time.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -93,6 +93,7 @@ RUN mkdir /app/django/src \
  && touch /app/django/src/rdwatch_scoring/__init__.py \
  && touch /app/django/README.md \
  && poetry install --with dev
+RUN git config --global --add safe.directory /app/django
 
 
 # Built static assets for vue-rdwatch
--- a/Dockerfile
+++ b/Dockerfile
@@ -84,6 +84,8 @@ RUN mkdir /app/django/src \
 FROM builder AS dev
 WORKDIR /app/django
 COPY django/pyproject.toml django/poetry.lock /app/django/
+# Copy git metadata to enable display of version information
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -43,6 +43,8 @@ CMD [ \
 # Base builder
 FROM base as builder
 COPY docker/keyrings/nodesource.gpg /usr/share/keyrings/nodesource.gpg
+# include .git directory to allow reading version tag from git
+COPY .git/ /app/
```
BryonLewis, line 47: The Vue version used it at build time when building the /dist.  Correct me if I'm wrong but the settings.py is executed at run-time meaning that you need to have a .git directory in the resulting end image folder or parent folder when running the application?

Have you tested the built image to see if the resulting end image has this directory and can read the tag information?

In addition to this it may be nice to add the .git info to the 'dev' tag so people running it can also quickly know what version they're on.

BryonLewis (changes requested):
I'm not entirely sure but I think there may be some changes needed to make sure the .git directory gets into the final image so that the subprocess command can run properly in a container?

Dockerfile
mvandenburgh: I fixed these issues and tested the built image, and everything works as expected for me

BryonLewis (approved):

PR #381, opened 2024-03-21
Move SAM model downloading to a task

BryonLewis:
TODO:
- [x] Add a volume to the celery container so it doesn't redownload the elements on docker compose down
- [x] How to handle the instance of trying to run SAM before the sam model is downloaded.  Do I attempt to download it if it isn't there or just wait until the model is downloaded?  I have a temporary fix but I think a better one would be to wait until the file exists.

I know the guard on Initiating SAM before downloading the file isn't perfect (although I do believe that `requests` will start creating the file before it's downloaded fully so at worse the SAM generation will fail when it tries to access the file).


django/src/rdwatch/tasks/__init__.py
```diff
@@ -641,3 +661,38 @@ def generate_image_embedding(id: int):
     except Exception as e:
         # Handle exceptions (e.g., logging, showing error messages)
         print(f'Error processing image {site_image.id}: {e}')
+
+
+@signals.worker_ready.connect
+def download_sam_model_if_not_exists(**kwargs):
+    file_path = '/data/SAM/sam_vit_h_4b8939.pth'
```
mvandenburgh, line 668: Can you make this a [`PathValue`](https://django-configurations.readthedocs.io/en/stable/values.html#configurations.values.PathValue) in settings.py instead? That way we a) have this defined in a single place, and b) someone running celery "natively" can choose a different location on disk for the pth file.

django/src/rdwatch/tasks/__init__.py
```diff
@@ -596,9 +598,27 @@ def generate_site_images_for_evaluation_run(
...
         checkpoint = '/data/SAM/sam_vit_h_4b8939.pth'
+        if not os.path.exists(
+            checkpoint
+        ):  # somehow the initial download didn't happen so we download the model again
+            #  This may download the file twice if the user attempts
+            # to do SAM before the initial download completes
+            # this is a temporary solution until we start reporting SAM model
+            # status in the /status endpoint and reflect this
+            # in the front-end to prevent SAM before the model is downloaded.
+            download_sam_model_if_not_exists()
```
mvandenburgh, line 614: I tested this locally, and it looks like celery will not pull any tasks off the queue until the `worker_ready` signal handler has finished executing. So, this file should always be available during this task, so I don't think this logic is needed. 
Did you observe the worker kicking off a task before the signal handler finished?

mvandenburgh (changes requested):

django/src/rdwatch/tasks/__init__.py
BryonLewis: I never directly observed it.  So I'll remove this, thanks for checking.

BryonLewis:
The `is_integer` wasn't going to work on the download progress because it was rare that the progress reported was exactly an integer.  I replaced it with some logic to only report on 5% increments for the downloading of the file.  Also thanks for reminding about running it locally and defining a location for the SAM model.

django/src/rdwatch/tasks/__init__.py
```diff
@@ -641,3 +643,48 @@ def generate_image_embedding(id: int):
...
+        try:
+            url = 'https://dl.fbaipublicfiles.com/segment_anything/sam_vit_h_4b8939.pth'
+            response = requests.get(url, stream=True)
+
+            # Check if the request was successful (status code 200)
+            if response.status_code == 200:
+                total_size = int(response.headers.get('content-length', 0))
+                bytes_downloaded = 0
+
+                with open(file_path, 'wb+') as file:
```
mvandenburgh, line 665:
```suggestion
                with open(file_path, 'wb') as file:
```

I'm not sure why I thought adding `+` here would solve anything - it turned out to be a different issue with my local system. Just using the `wb` mode should be fine, as the `+` just allows read access as well (which we don't need here and we're just writing the file). Sorry for the confusion :sweat_smile: 

(`open` docs for reference: https://docs.python.org/3/library/functions.html#open)

mvandenburgh (approved):
LGTM after the previously mentioned revert. I pulled and tested and it works :+1:

mvandenburgh:
@BryonLewis I'm going to merge this so I can ensure we get the image building in time for the release tomorrow.

PR #383, opened 2024-03-25
Dev git hash

BryonLewis:
Wanted the version to show up in the standard docker compose up.  So added it to the docker compose override and it also required adding a git command to the 'dev' target of the dockerfile.


mvandenburgh (approved):
Ah, it was appearing for me in docker compose up, but I don't usually use that to run it so I might be missing something

PR #384, opened 2024-03-26
Add year filter to vector tiles endpoint

mvandenburgh:
Adds a year filter to the vector tiles endpoint. This makes it so that the client must supply a year representing the range of site observations they wish to view. The client makes additional requests if the user slides the time slider to a new year.

This resolves the performance issues for large sites that contain a significant amount of observation polygons.

### DIFF
--- a/tileserver/devServer.mjs
+++ b/tileserver/devServer.mjs
@@ -7,9 +7,9 @@ const fastify = Fastify({ logger: true });
 await fastify.register(cors);
 
 fastify.get('/api/vector-tiles/tile.pbf', async (request, reply) => {
-  const { modelRunId, z, x, y, randomKey } = request.query;
+  const { modelRunId, z, x, y, randomKey, year } = request.query;
 
-  const vectorTileData = await getRGDVectorTiles(modelRunId, z, x, y, randomKey);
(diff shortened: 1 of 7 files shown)


BryonLewis (commented):
There is one thing I forgot about when mentioning doing the Year by Year for the site Observations.  The observations are a single point in the time but we display them as if they are persistent after they occur. Sites themselves have a time range ( a beginning and end date) so we display them for the duration of the site and let them persist afterwards so that it will show all the sites.  This is to solve the issue of going to present day and not being able to see a specific site because it was only active from 20180305 to 201800308.

So how Site observation displaying works:
- if we have an observation on December 15th 2019 it stays displayed as the state until the next observation occurs or until the site time limits are reached.

Now the problem occurs when filtering by year we eliminate any visualization of sites until the first observation for the year:
- If the previous observation for a site is on December 28th and the next Observation is on January 20th of the next year, as soon as we switch to a new year we don't see any observations until the 20th.  Even though technically we would want that previous observation to persist.

So one possible solution is to modify the results of the query/filter so it grabs the last sorted timestamp observation from before the year as well.  I.E it will add to the next year's vector-tiles the last observation before the currently selected year.

BryonLewis (approved):
Nice, just tested across proposals/scoring and the default view and it seems to work properly.

PR #387, opened 2024-03-27
Switch to cpu-only version of pytorch, fix poetry/tox integration

mvandenburgh:
@BryonLewis can you test this locally and see if it works for you? It appears to work for me, but I'm never sure if my docker container is actually reinstalling dependencies or using some cache.


mvandenburgh:
Ah, I see what you meant by tox not working with it now.

BryonLewis (approved):
thanks for doing this.  I had just stopped before once tox was having a issue.

PR #391, opened 2024-04-03
UI Upgrade

### LINKED ISSUE #389: UI Upgrading
Figma Link: https://www.figma.com/file/HzfMOKyPaKVVDN7OEKGfgs/SMART-phase-2?type=design&node-id=0-1&mode=design

Overall Process:
- Restructure the main ModelRun listing with the new styles and new icons
- Restructure the site model listing for proposals
- Adjust the legend location and make it dismissable

Image Viewer Editor Redesign:
Separated out because I think it may be one of the bigger processes:
- Reconfigure so it isn't a dialog it is either in a new screen or on the same screen
- After making it a global dialog reconfigure to align everything in the left/right.

BryonLewis:
resolves #389

### DIFF
--- a/vue/src/components/LayerSelection.vue
+++ b/vue/src/components/LayerSelection.vue
@@ -153,23 +153,25 @@ const toggleScoring = (data? : undefined | 'simple' | 'detailed') => {
           class="px-2 mx-2"
           v-bind="props"
           size="large"
-          :color="state.filters.drawObservations ? 'rgb(37, 99, 235)' : ''"
+          :color="state.filters.drawObservations ? 'primary' : ''"
           style="min-width: 200px"
           @click="toggleObs()"
         >
           <span>
             Observations
           </span>
-          <v-icon
+          <div
             v-if="state.filters.drawObservations?.includes('model')"
+            class="layer-icon mx-1"
           >
-            mdi-alpha-m-box
-          </v-icon>
-          <v-icon
+            M
+          </div>
+          <div
             v-if="state.filters.drawObservations?.includes('groundtruth')"
+            class="layer-icon mx-1"
           >
-            mdi-alpha-t-box
-          </v-icon>
+            G
+          </div>
         </v-btn>
       </template>
       <v-card outlined>
@@ -178,19 +180,33 @@ const toggleScoring = (data? : undefined | 'simple' | 'detailed') => {
             value="Model"
             @click="toggleObs('model')"
           >
-            Model
-            <v-icon>
-              mdi-alpha-m-box
-            </v-icon>
+            <v-btn
+              size="large"
+              variant="text"
+            >
+              Model
+              <div
+                class="layer-icon mx-1"
+              >
+                M
+              </div>
+            </v-btn>
           </v-list-item>
           <v-list-item
             value="GroundTruth"
             @click="toggleObs('groundtruth')"
           >
-            GroundTruth
-            <v-icon>
-              mdi-alpha-t-box
-            </v-icon>
+            <v-btn
+              size="large"
+              variant="text"
+            >
+              GroundTruth
+              <div
+                class="layer-icon mx-1"
+              >
+                G
+              </div>
+            </v-btn>
           </v-list-item>
         </v-list>
       </v-card>
@@ -203,28 +219,32 @@ const toggleScoring = (data? : undefined | 'simple' | 'detailed') => {
           class="px-2 mx-2"
           v-bind="props"
           size="large"
-          :color="state.filters.drawSiteOutline ? 'rgb(37, 99, 235)' : ''"
+          :color="state.filters.drawSiteOutline ? 'primary' : ''"
           style="min-width:150px"
           @click="toggleSite()"
         >
           <span>
             Sites
           </span>
-          <v-icon
+          <div
             v-if="state.filters.drawSiteOutline?.includes('model')"
+            class="layer-icon mx-1"
           >
-            mdi-alpha-m-box
-          </v-icon>
-          <v-icon
+            M
+          </div>
+          <div
             v-if="state.filters.drawSiteOutline?.includes('groundtruth')"
+            class="layer-icon mx-1"
           >
-            mdi-alpha-t-box
-          </v-icon>
-          <v-icon
+            G
+          </div>
+          <div
             v-if="state.filters.siteTimeLimits"
+            size="x-large"
+            class="layer-icon mx-1"
           >
-            mdi-clock
-          </v-icon>
+            <v-icon>mdi-clock </v-icon>
+          </div>
         </v-btn>
       </template>
       <v-card outlined>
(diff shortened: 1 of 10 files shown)


django/src/rdwatch/views/model_run.py
```diff
@@ -469,10 +497,27 @@ def get_proposals_query(model_run_id: UUID4):
 
 @router.get('/{model_run_id}/proposals/')
 def get_proposals(request: HttpRequest, model_run_id: UUID4):
-    region = get_object_or_404(ModelRun, pk=model_run_id).region
+    data = get_model_run_details(model_run_id).values_list('json', flat=True)
+    if not data.exists():
+        raise Http404()
+
+    query = get_sites_query(model_run_id)
+    model_run = data[0]
```
mvandenburgh, line 505:
```suggestion
    model_run = data.first()
```

django/src/rdwatch/views/satellite_fetching.py
```diff
@@ -0,0 +1,26 @@
...
+):
+    running_sites = SatelliteFetching.objects.filter(
+        status=SatelliteFetching.Status.RUNNING
+    )
+    if model_runs is not None and len(model_runs) > 0:
+        running_sites = running_sites.select_related('sites').filter(
+            site__configuration__in=model_runs
+        )
+    running_site_ids = list(running_sites.values_list('site_id', flat=True))
+    return running_site_ids
```
mvandenburgh, line 26:
```suggestion
router = RouterPaginated()
logger = logging.getLogger(__name__)


@router.get('/running/')
def satellite_fetching_running(
    request: HttpRequest, model_runs: list[UUID4] = Query(None)  # noqa: B008
):
    running_sites = SatelliteFetching.objects.filter(
        status=SatelliteFetching.Status.RUNNING
    )
    if model_runs is not None and len(model_runs) > 0:
        running_sites = running_sites.filter(
            site__configuration__in=model_runs
        )
    running_site_ids = running_sites.values_list('site_id', flat=True)
    return running_site_ids
```

Two things -
1) I don't think the select_related is necessary
2) Not paginating this endpoint + casting the full queryset to a list could result in excessive memory usage if there are thousands of satellite fetchings all at once. I don't think that would ever realistically happened outside of someone maliciously trying to crash the server, but paginating it is simple enough that it seems like it'd be worth it?

django/src/rdwatch_scoring/api.py
```diff
@@ -12,6 +12,7 @@
 api.add_router('/regions/', views.region.router)
 api.add_router('/sites/', views.site.router)
 api.add_router('/eval-numbers/', views.eval.router)
+api.add_router('/satellite_fetching/', views.satellite_fetching.router)
```
mvandenburgh, line 15:
```suggestion
api.add_router('/satellite-fetching/', views.satellite_fetching.router)
```
Same as above

mvandenburgh (changes requested):
Some preliminary comments/review on the backend code. Next, I'll take a pass at the frontend code, and then I'll pull the branch down and test it out in a final review

mvandenburgh:
I noticed something strange with the checkboxes on sites when testing this. Selecting a site by clicking it on the map doesn't fill the checkbox -
Peek 2024-05-01 12-41

I believe the checkbox should get filled here? Unless I'm misunderstanding the UX

BryonLewis:
> I believe the checkbox should get filled here? Unless I'm misunderstanding the UX

I had that on in the beginning and I don't remember why I turned it off.  Give me some time, I can easily turn it back on.

BryonLewis:
> I believe the checkbox should get filled here? Unless I'm misunderstanding the UX

I redid the selection logic for the regular RGD viewer and the annotation viewer.
I also fixed the scrolling tool.  It was working before I switched over to utilize virtual scroll lists.  Added code to handle this so control + hover over sites will now scroll to the site in the list that is being hovered over.

mvandenburgh:
Frontend code changes and the proposal view lgtm. Only thing I haven't reviewed/tested yet is the scoring app backend changes, my local scoring db got messed up so I have to repopulate that. I'll do that tomorrow and get a final review posted then.

django/src/rdwatch_scoring/views/satellite_fetching.py
```diff
@@ -0,0 +1,24 @@
...
+def satellite_fetching_running(
+    request: HttpRequest, model_runs: list[UUID4] = Query(None)  # noqa: B008
+):
+    running_sites = SatelliteFetching.objects.filter(
+        status=SatelliteFetching.Status.RUNNING
+    )
+    if model_runs is not None and len(model_runs) > 0:
+        running_sites = running_sites.filter(model_run_uuid__in=model_runs)
+    running_site_ids = list(running_sites.values_list('site', flat=True))
+    return running_site_ids
```
mvandenburgh, line 24: This also needs to be updated as the `rdwatch` app endpoint was

vue/src/client/services/ApiService.ts
```diff
@@ -162,6 +173,24 @@ export class ApiService {
...
+   * @param id
+   * @returns SiteObservationList
+   * @throws ApiError
+   */
+    public static getSite(
+      id: string
+    ): CancelablePromise<SiteInfo> {
+      return __request(OpenAPI, {
+        method: "GET",
+        url: `${this.getApiPrefix()}/sites/{id}`,
```
mvandenburgh, line 186:
```suggestion
        url: `${this.getApiPrefix()}/sites/{id}/`,
```

django/src/rdwatch/views/site_image.py
```diff
@@ -23,7 +23,7 @@ class SiteImageSchema(Schema):
     source: str
     cloudcover: float | None
     image: str
-    observation_id: UUID4 | None
+    observation_id: UUID4 | str | None
```
mvandenburgh, line 26: Leftover from debugging?
```suggestion
    observation_id: UUID4 | None
```

mvandenburgh (changes requested):

django/src/rdwatch/views/site_image.py
BryonLewis: Yup

django/src/rdwatch_scoring/views/satellite_fetching.py
BryonLewis: I'm assuming you mean the Pagination and the response type?

django/src/rdwatch_scoring/views/satellite_fetching.py
```diff
@@ -0,0 +1,26 @@
...
+@paginate(LimitOffsetPagination)
+def satellite_fetching_running(
+    request: HttpRequest, model_runs: list[UUID4] = Query(None)  # noqa: B008
+):
+    running_sites = SatelliteFetching.objects.filter(
+        status=SatelliteFetching.Status.RUNNING
+    )
+    if model_runs is not None and len(model_runs) > 0:
+        running_sites = running_sites.filter(model_run_uuid__in=model_runs)
+    running_site_ids = list(running_sites.values_list('site', flat=True))
```
mvandenburgh, line 25:
```suggestion
    running_site_ids = running_sites.values_list('site', flat=True)
```

Coercing to a list negates the benefits of pagination here

mvandenburgh (approved):
Just one last change needed to the scoring satellite fetch endpoint. Once that's addressed, I think this is good to go :+1:

django/src/rdwatch_scoring/views/satellite_fetching.py
BryonLewis: scattterbrained today.

PR #396, opened 2024-04-11
Associate regions with ModelRuns instead of SiteEvaluations

mvandenburgh:


mvandenburgh:
Ah, I forgot about the node vector tile server, that needs to be updated too.

django/src/rdwatch/models/site_evaluation.py
```diff
@@ -179,10 +170,6 @@ def bulk_create_from_region_model(
 
         site_evals: list[SiteEvaluation] = []
         with transaction.atomic():
-            region = get_or_create_region(
-                region_feature.properties.region_id, region_feature.parsed_geometry
-            )[0]
-
```
BryonLewis, line 185: Okay, sorry but bear with me on this because it may take a while and I may also be wrong.

So This is being removed because we don't need to create a region each time we create a new site.  The probably is I don't believe we have a spot where we upload the region geometry anymore to the Region model.  This was done here with the region_feature.parse_geometry.  It was done a single time, before iterating over all of the other items in the list.


I went looking and noticed that this `bulk_create_from_region_model` is only a  an endpoint located here: https://github.com/ResonantGeoData/RD-WATCH/blob/model-run-region/django/src/rdwatch/views/model_run.py#L375-L379

This is called when loading groundTruth through the script.  It will use this endpoint and I don't believe there is a way it will create the region anymore:
https://github.com/ResonantGeoData/RD-WATCH/blob/model-run-region/scripts/loadGroundTruth.py#L96

I think just enabling this line will create that new geometry.


SideNote:  I don't know if we want to region-model to be adding all these site evaluations at the same time?  Should it be its own contained thing that just creates the region model for the system instead of creating these Sites as well?

BryonLewis (changes requested):
Mostly the tile server functionality.
Then making it so you can add new region polygons into the system.  I didn't test fully but it looks like there is no way to add new data to the `region.geom` anymore.

django/src/rdwatch/models/site_evaluation.py
mvandenburgh: Oh yeah you're right, good catch! Fixed

django/src/rdwatch/views/vector_tile.py
```diff
@@ -115,7 +115,7 @@ def vector_tile(request: HttpRequest, model_run_id: UUID4, z: int, x: int, y: in
                 timemax=ExtractEpoch('end_date'),
                 performer_id=F('configuration__performer_id'),
                 performer_name=F('configuration__performer__slug'),
-                region=F('region__name'),
+                region=F('configuration__region__name'),
```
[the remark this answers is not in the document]
mvandenburgh: Updated https://github.com/ResonantGeoData/RD-WATCH/pull/396/commits/6cf9119781dc31b477c49dc8d9874b824523e178

django/src/rdwatch/migrations/0026_remove_siteevaluation_region_modelrun_region.py
```diff
@@ -0,0 +1,25 @@
...
+            name='region',
+            field=models.ForeignKey(
+                help_text='The region this model run belongs to',
+                null=True,
+                on_delete=django.db.models.deletion.PROTECT,
+                to='rdwatch.region',
+                related_name='model_runs',
+            ),
+        ),
+    ]
```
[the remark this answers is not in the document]
mvandenburgh: Like i mentioned offline, we need at least two. I was able to eliminate the third one though https://github.com/ResonantGeoData/RD-WATCH/pull/396/commits/28150e9c150691f0ed05abe850b3124f6fc272c9.

mvandenburgh:
@BryonLewis I believe I addressed all of your requested changes. I also wanted to mention, I did a minor refactor of the vector tileserver to use the `postgres.js` library instead of `node-pg` https://github.com/ResonantGeoData/RD-WATCH/pull/396/commits/1a8fa2f5acddc7850b126186a61f817d0287aa1d. That library makes it way easier to pass parameters into queries, as you can use a JS template literal with string interpolation instead of passing an array and referencing params by index. This should make updating these queries less of a hassle in the future.

tileserver/rgd.mjs
```diff
@@ -13,270 +11,271 @@ const logging = process.env.RDWATCH_VECTOR_TILE_LOGGING || false;
...
+        "rdwatch_siteevaluation"."status",
+        "rdwatch_siteevaluation"."cache_originator_file",
+        "rdwatch_siteevaluation"."cache_timestamp",
+        "rdwatch_siteevaluation"."cache_commit_hash",
+        "rdwatch_siteevaluation"."id" AS "id",
+        "rdwatch_siteevaluation"."id" AS "uuid",
+        ST_AsMVTGeom(
           "rdwatch_siteevaluation"."geom",
-          ST_TileEnvelope($1, $2, $3)
+          ST_TileEnvelope(${z}, ${x}, ${y})
```
mvandenburgh, line 38: This _looks like_ it's vulnerable to SQL injection, but `postgres.js` actually sanitizes these (see [docs](https://github.com/porsager/postgres#query-parameters)) via the `sql` template literal.

BryonLewis: nice, a lot better than continually looking up the params.

django/src/rdwatch/models/site_evaluation.py
```diff
@@ -179,9 +171,12 @@ def bulk_create_from_region_model(
...
         site_evals: list[SiteEvaluation] = []
         with transaction.atomic():
-            region = get_or_create_region(
-                region_feature.properties.region_id, region_feature.parsed_geometry
-            )[0]
+            # Update the region of the configuration.
+            # If the region doens't have a geometry,
+            # the one from the region model is used.
+            ModelRun.objects.filter(pk=configuration.pk).update(
+                region=get_or_create_region(region_feature.properties.region_id)[0]
```
BryonLewis, line 178: I think you need to add the `region_feature.parsed_geometry as the second argument to the `get_or_create_region` function

BryonLewis (changes requested):
Thanks for the updates.
I have request change to include the geometry when creating new regions.

I also have a PR to fix some proposal and scoring references to regions.

django/src/rdwatch/models/site_evaluation.py
mvandenburgh: :facepalm: fixed https://github.com/ResonantGeoData/RD-WATCH/pull/396/commits/e556ca4aa6a2ec7d5e72196fb94eb5c0d02ea5a4

BryonLewis (approved):
👍

PR #392, opened 2024-04-09
Proposals for T&E scoring db

JHUAPL-hjg1:

PR #400, opened 2024-04-18
Proposals scoring review

mvandenburgh:
This PR is set to merge into #392 and serves as the rest of my review for that PR. For the most part it makes changes such as using convenience methods provided by Django/Ninja such as `get_object_or_404`, or adding query params for `GET` endpoints as function arguements instead of pulling them out of the `request` object ([relevant docs for that](https://django-ninja.dev/guides/input/query-params/)).

I haven't looked at any of the vector tile code yet, I will do that next and port them to the node server.

### DIFF
--- a/django/src/rdwatch_scoring/router.py
+++ b/django/src/rdwatch_scoring/router.py
@@ -31,7 +31,7 @@ class ScoringRouter:
         """
         if model in RGD_DB_MODELS or model._meta.app_label == 'rdwatch':
             return 'default'
-        return None
+        return 'scoringdb'
 
     def allow_relation(
         self, obj1: type[Model], obj2: type[Model], **hints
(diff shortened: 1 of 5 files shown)


django/src/rdwatch_scoring/router.py
```diff
@@ -31,7 +31,7 @@ def db_for_write(self, model: type[Model], **hints) -> DbName | None:
         """
         if model in RGD_DB_MODELS or model._meta.app_label == 'rdwatch':
             return 'default'
-        return None
+        return 'scoringdb'
```
BryonLewis, line 34: Could we reverse this to be excplicit when `model._meta.app_label === 'rdwatch_scoring'` like the db_for_read about  I would want the default to be the 'default'

BryonLewis (changes requested):
One change about default db model.

BryonLewis (approved):

PR #403, opened 2024-04-24
Add index to `SatelliteFetching.status` column

mvandenburgh:
I tested this on the production database, and this brings the non-cached response time for the model run endpoint from 19 seconds to about 3.


BryonLewis (approved):

BryonLewis:
Wait, should this be added as anindex for the Scoring version as well?  I mean adding it to BaseSatelliteFetching instead of just SatelliteFetching, because we have a version of SatelliteFetching in the Scoring DB ?

PR #404, opened 2024-04-24
Auto delete old satellite fetches

mvandenburgh:
This PR consolidates the two existing "garbage collection" tasks (cleaning up old model runs + annotation exports) into one task, and also adds scheduled cleanup of Satellite fetches that are over a day old.

### DIFF
--- a/django/src/rdwatch/tasks/__init__.py
+++ b/django/src/rdwatch/tasks/__init__.py
@@ -393,7 +393,7 @@ def get_siteobservations_images(
 
 
 @shared_task
-def delete_temp_model_runs_task() -> None:
+def collect_garbage_task() -> None:
     """Delete all model runs that are due to be deleted."""
     model_runs_to_delete = (
         ModelRun.objects.filter(expiration_time__isnull=False)
--- a/django/src/rdwatch/tasks/__init__.py
+++ b/django/src/rdwatch/tasks/__init__.py
@@ -443,6 +443,11 @@ def collect_garbage_task() -> None:
         created__lte=timezone.now() - timedelta(hours=1)
     ).delete()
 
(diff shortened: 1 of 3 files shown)


django/src/rdwatch/tasks/__init__.py
```diff
@@ -438,14 +438,15 @@ def delete_temp_model_runs_task() -> None:
...
+    # Delete all S3 Export Files that are over an hour old
+    AnnotationExport.objects.filter(
         created__lte=timezone.now() - timedelta(hours=1)
-    )
-    exports_to_delete.delete()
+    ).delete()
+
+    # Delete all SatelliteFetching tasks that are over an day old
+    SatelliteFetching.objects.filter(
+        timestamp__lte=timezone.now() - timedelta(days=1)
```
BryonLewis, line 448: may want a status filter on this so they are not in the status of "Running", I.E. they are Complete or Error.  Just so if we have some insane long-running download tasks that are in status running it doesn't get removed for no reason.  It's a rare corner case but could happen.

BryonLewis (changes requested):
add status filter the SatelliteFetching deletion

django/src/rdwatch/tasks/__init__.py
mvandenburgh: I was hesitant to add this because a task that dies to to a bug for example, may never set its SatelliteFetching to Complete or Error. But I think I found a good alernative, I changed it to check the associated celery `TaskResult` status, and if that is still `Running` then it doesn't get deleted.

BryonLewis: perfect, that's a better implementation.

BryonLewis (approved):

PR #407, opened 2024-04-26
Rework `Performer` model, allow creation of new performers

mvandenburgh:
This PR does the following -

- Seperates the `Performer` model from `lookups.py` into its own file
- Renames the fields on the model to be clearer
- Simplifies the performer filter logic on the model runs listing endpoint
- Adds an endpoint for creating new performers + a test for that endpoint and existing ones

This will enable users to create their own performers outside of the defaults provided in the lookups fixture.


BryonLewis (approved):
This looks good I just have one question regarding workflow.  I think this can probably be merged either way but my question may result in an additional PR.

Question:
Should we make it so when uploading sites that have a built-in performer that is in their JSON but not in the database that they automatically add the new performer?
I.E if we had a site with performer 'ABC' it would just add a new performer with the shortcode 'ABC' and the description 'ABC'.  Or should we have it error when that happens (like it does now)?

PR #417, opened 2024-05-08
Fix 500 error in site image endpoint

mvandenburgh:
Fixes https://smartgitlab.com/TE/RD-WATCH/-/issues/103#note_44321.


BryonLewis (approved):
Thanks for finding the issue.
I think we can handle this case, but should we also correct the model to enable null on default like the non-scoring `SiteImage`?
https://github.com/ResonantGeoData/RD-WATCH/blob/main/django/src/rdwatch/models/site_image.py#L77
I think it was defaulting to `''` instead of `null/None` when no value was provided in the task.
https://github.com/ResonantGeoData/RD-WATCH/blob/main/django/src/rdwatch_scoring/models/site_image.py#L9

mvandenburgh:
> I think we can handle this case, but should we also correct the model to enable null on default like the non-scoring `SiteImage`? [`main`/django/src/rdwatch/models/site_image.py#L77](https://github.com/ResonantGeoData/RD-WATCH/blob/main/django/src/rdwatch/models/site_image.py?rgh-link-date=2024-05-08T16%3A34%3A59Z#L77) I think it was defaulting to `''` instead of `null/None` when no value was provided in the task. [`main`/django/src/rdwatch_scoring/models/site_image.py#L9](https://github.com/ResonantGeoData/RD-WATCH/blob/main/django/src/rdwatch_scoring/models/site_image.py?rgh-link-date=2024-05-08T16%3A34%3A59Z#L9)

Yeah, that's exactly what is happening, it's returning `''` for the observation_id. I think the better solution would be to add something on the ninja response schema to coerce `''` into `None`, since it's bad practice in Django to allow a nullable CharField (https://docs.astral.sh/ruff/rules/django-nullable-model-string-field/). 

I can make a follow up PR for that, in the meantime I just merged this to fix the issue

PR #421, opened 2024-05-10
Use `poetry-dynamic-versioning` plugin for versioning

mvandenburgh:


BryonLewis (approved):
pulled and checked locally, hopefully this works for the deployed version.

PR #422, opened 2024-05-10
GroundTruth Refactor

### LINKED ISSUE #418: GroundTruth Interaction Update
So ground truth in the RD-WATCH DB is a flag on the model run itself.

Info:
- Want the ability to be able to toggle on/off ground truth for any items in RD-WATCH without having to change significantly other models
- GroundTruth are a model run that are loaded with the GT flag on.
- Currently it will search the current list of items and toggle on the GT if is available when you click on the GT item

Possible Updates:
- Refactor the searching for model-Runs so that it always include the GT for a region and sorted near the top of the list
- The GT is by default hidden from the system by a flag (hide GT) that will not show it.  Maybe a warning/symbol if there is no GT for the region
- When a specific GT flag is toggled on in the interface it makes it so the Site defaults to showing only the GT.  Right now there is some weird behavior between toggling on/off the model-run sites vs the GT sites.

Complications:
- RDWATCH Scoring DB uses a different system o toggle GT on/off.

BryonLewis:
resolves #418 

- May be a slight performance penalty but I think it's better to have non-groundtruth provide a direct link to the ground truth so it can be used to toggle on/off.
- The linked groundtruth will be the matching region with `ground_truth=True` with the newest date.
- The ground truth is not shown by default and instead in RDWatch it will link to ground truth when you turn it on using the Layer buttons.
- The groundtruth and model for the layer selections will enable/disable automatically based on the selected model.  So groundtruth is inherently connected to the model-run instead of having to be toggled automatically.
- Updated the scoring response to label ground-truth vs the regular items respectively in the list.
- Updated the regular RDWatch response to label groundtruth vs the regular items as well.

https://github.com/ResonantGeoData/RD-WATCH/assets/61746913/ac762d5a-fce1-49d7-b502-b7ee5e1c0c4c

### DIFF
--- a/vue/src/components/LayerSelection.vue
+++ b/vue/src/components/LayerSelection.vue
@@ -39,10 +39,10 @@ const groundTruthState = computed<GroundTruthState>(() => {
     }
   });
   if (hasGroundTruth) {
-    result = 'HasGroundTruth';
+    result = GroundTruthState.HasGroundTruth;
   }
   // Check to see if it is ground Truth
-  if (result === 'HasGroundTruth') {
+  if (result === GroundTruthState.HasGroundTruth) {
     const keys = Object.keys(state.groundTruthLinks);
     const keysLength = keys.length;
     let selfReferencedKeys = 0;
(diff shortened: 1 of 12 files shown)


vue/src/components/LayerSelection.vue
```diff
@@ -11,6 +11,56 @@ const route = useRoute();
...
+  let result: 'NoGroundTruth'|'HasGroundTruth'|'AllGroundTruth'  = 'NoGroundTruth';
+  if (scoringApp.value) {
+    return 'HasGroundTruth';
+  }
+  if (proposals.value) {
+    if (state.proposals.ground_truths) {
+      return 'HasGroundTruth';
+    } else {
+      return 'NoGroundTruth';
+    }
```
mvandenburgh, line 27: Would be cleaner to use an enum here, imo (not that this is as big a deal as if it were python, as the TS compiler would catch any string literal typos anyway, but I think it's slightly cleaner).

```suggestion
enum GroundTruthState {
  NoGroundTruth = 0,
  HasGroundTruth = 1,
  AllGroundTruth = 2,
}

// Either has ground truth, is ground truth, or no ground truth
const groundTruthState = computed<GroundTruthState>(() => {
  let result: GroundTruthState  = GroundTruthState.NoGroundTruth;
  if (scoringApp.value) {
    return GroundTruthState.HasGroundTruth;
  }
  if (proposals.value) {
    if (state.proposals.ground_truths) {
      return GroundTruthState.HasGroundTruth;
    } else {
      return GroundTruthState.NoGroundTruth;
    }
```

(Would also need to update rest of the file)

mvandenburgh (commented):
Just left some code suggestions. Otherwise LGTM

mvandenburgh (approved):

PR #424, opened 2024-05-14
Misc fixes/improvements to ingest endpoint

mvandenburgh:
This PR fixes some issues I noticed while debugging the bug David reported about 500 errors on model run upload.

- After #407, new performers can be created. But, the pydantic validation for site/region models still restricts the list of accepted performers. 
- If an invalid geojson polygon is provided, GeoDjango throws a `GEOSException`. We don't explicitly catch it currently, resulting in a 500 error.


BryonLewis (approved):
Sorry I thought I had approved this last week.

PR #428, opened 2024-05-20
Creating Site Proposals

### LINKED ISSUE #427: Generate new Site Proposal
A way to create a new site proposal at any location from nothing.  In the future this may be modified to copy from existing proposals but lets start with the basics.

A proposal SitecConsists of:

- Region
- SiteId
- Geometry,
- Start/End Times (which can be null)
- A Label

Process:

- User clicks on some button to add a new proposal
- Kicks off the system into proposal creation mode
- This mode allows them to draw a polygon
- After drawing the polygon they then need to set a label and a start/end date which can be null
- Finally submission goes to an endpoint and pushes up all the data for the newly proposed site.

BryonLewis:
resolves #427 
- Adds button in proposal mode to create Site Proposals
- When clicked it opens a side panel that has instructions about drawing a polygon
- User can draw the polygon and then edit the metadata such as time, label, comments and siteId number.
- Added in the ability to turn on regional satellite images for proposal view for drawing new annotations.
- Updated the proposal view to add in the satellite timestamps so that a satelliteview can be used when drawing new proposals.

https://github.com/ResonantGeoData/RD-WATCH/assets/61746913/473f6d4d-a27d-4cbe-9ff2-0f0704e718c6


mvandenburgh (approved):
LGTM, pulled and tested locally and it works for me :+1:

PR #430, opened 2024-05-21
Switch to Redis for task queue

mvandenburgh:
This PR configures celery to use the existing redis instance (the one used for caching) for queueing instead of provisioning a separate RabbitMQ instance.

@BryonLewis I applied this change to the rgd-dev instance as well, if you'd like to verify that as well.


mvandenburgh:
GH actions is having some issues it seems https://www.githubstatus.com/

BryonLewis (changes requested):
I'm guessing these might need some updating as well?

I don't believe we do any task testing or start a rabbit instance but probably should update it anyways?
https://github.com/ResonantGeoData/RD-WATCH/blob/redis-queue/.github/workflows/ci.yaml#L14

For your native dev:
https://github.com/ResonantGeoData/RD-WATCH/blob/redis-queue/dev/.env.docker-compose-native#L12

It seemed to work for both image downloading and site downloading (multiple sites into a Zip file) while in the docker mode.

BryonLewis (approved):

PR #433, opened 2024-05-24
UI Upgrade Fixes

### LINKED ISSUE #429: Additional UI Refactor Comments
https://www.figma.com/design/HzfMOKyPaKVVDN7OEKGfgs/SMART-phase-2?node-id=417-6531&t=FPGifbUcfw6ElDu4-0

- Add buttons to right align
- Revert search bar to a true search bar
- Make the number of sites be in a parenthesis in 
- Add collapse button
- Add the Deselect All button for selected sites
- Change the no images downloaded to an icon-badge.
- Switch it back to the whole card instead of the checkbox instead.
- swap the slider to be above the buttons
- In the selected color the color border is wrong and make it so the yellow color goes around the full size.  So remove the corners for the selected sites.

Look into making it so the Tooltip when the mouse is over it, it doesn't close automatically.  It only closes after the user mouses out of the area.

### LINKED ISSUE #440: Create Client Only Post site-model/site-region endpoints
Relates to #436 

Swap over to a non-CRSF exempt endpoint for the client to post site-models and site-regions.  In the future this may be changed if we find a good way to enable both authentications.

BryonLewis:
resolves #429 
resolves #440 

image

Fixes some various discrepancies between the mock-up and implementation.  Also adds some new features for selection/deselection.

Notes:

- utilized the Vite Environment variables to prevent redirection when in dev mode (running locally) so it doesn't redirect to a gitlab authentication
- Added in a modification to the Base Configuration for the `LOGIN_REQUIRED_IGNORE_PATHS` to add the `/api/` path when running in the developmentMode.  Without this you would need to login as the admin user to be able to see anything.

Done:

- Reorder the Slider and the buttons for the Model Run cards
- Update Model-Run Cards Radio button in proposal mode to be clicked properly
- Added in ability to click on Model-Run Card, Site Cards independently of clicking the checkbox/radio button
- Updated selection logic inside of the SiteList and SiteCard to clean it up and rely a bit more on the global `state.selectedSites`
- Added `@click.stop` to various components inside of the card to prevent deselecting when browsing images
- ToolTip has been refactored so that the system has the display settings within the tooltip Cog wheel instead of the Site filter dialog.  User can edit those filters by hovering over and utilizing the checkboxes inside of that area instead.
- Fixed a bug with the `randomKey` with the new endpoints for the Django Server.


vue/src/App.vue
```diff
@@ -8,11 +8,15 @@ const isLoggedIn = ref(false);
 // Check if the user is logged in before rendering any UI. If they're not,
 // redirect them to the login page.
 onBeforeMount(async () => {
-  try {
-    await fetch('/api/status/', { redirect: 'error' });
+  if (import.meta.env.PROD) {
```
mvandenburgh, line 11: Because Vite env vars are set at docker image build time, I don't think this will work.

BryonLewis: I think you're new authentication for the POST modelrun/sitemodel endpoints makes this moot anyways now.  I'll adjust this and the server settings.

mvandenburgh (approved):
pulled and tested LGTM :+1:

PR #436, opened 2024-06-03
Allow API key authentication on model run ingest endpoints

mvandenburgh:
This allows the user to authenticate with an API key instead of GitLab oauth on the three endpoints used to create a model run. I've also updated the `scripts/` upload scripts to use this new approach. Currently, the API key is just a global Django setting that we can set; it's really only intended to be used by the SmartFlow upload job.

Note, I started out using user-based HTTP Basic Auth, but that didn't work because it would require the user to retrieve a csrf token before making their request. Ideally we don't require a csrf token, but disabling Django's CSRF protection when using basic auth is dangerous because the `Authorization` header used in basic auth is treated the same way as cookies are (i.e., the browser sends it along with any request to the RD-WATCH domain automatically), so it would be vulnerbale to a CSRF attack.
Using a non-standard header (`X-RDWATCH-API-KEY` in our case) allows us to safely disable CSRF protection on these three endpoints, as the browser will not send it automatically on every request.


scripts/loadGroundTruth.py
```diff
@@ -24,10 +24,10 @@ def main():
         '--skip_regions', action='store_true', help='expiration time in hours'
     )
     parser.add_argument(
-        '--rgd-auth-cookie',
-        required=False,
+        '--rgd-api-key',
+        required=True,
         type=str,
-        help='RGD Authentication cookie, e.g.: token=<LONG BASE64 STRING>',
+        help='RGD API key',
```
BryonLewis, line 30: This is probably me being lazy but can we default to 'secretKey; it being the default for local dev.  For both this and the regular modelRun loading.

BryonLewis (changes requested):
Small request to default to 'secretKey'
Additional request to utilize multiple Auth for the site_model endpoint for adding new proposals.

django/src/rdwatch/views/model_run.py
```diff
@@ -379,7 +396,10 @@ def get_model_run(request: HttpRequest, id: UUID4):
 @router.post(
     '/{model_run_id}/site-model/',
     response={201: UUID4},
+    auth=[ModelRunAuth()],
```
BryonLewis, line 399: For some reason it didn't add my other comment.  I use this endpoint in the client for drawing new proposal sites.  So I think we need multiple authentication here:
https://django-ninja.dev/guides/authentication/?h=multiple+auth#multiple-authenticators
That way it can try using the django_auth through gitlab first and then fallback to the API header.

mvandenburgh: To do this, we need to make the CSRF exemption conditional on whether or not the API key auth or the default Django session auth is used. Since disabling CSRF protection for the default auth flow is insecure. I think we can use a custom decorator to do this, I'll look into it.

mvandenburgh: Deferring this to another PR as we discussed

BryonLewis (approved):
I added a new issue #440. I can do it at some point to at least temporarily add in new endpoints for the client site-model and region-model POST endpoints.  We can revisit in the future if there is a way to enable both forms of auth.

PR #438, opened 2024-06-04
Restructure application to follow Django/Resonant best practices

mvandenburgh:
Notable changes/improvements:

- Global `urls` and `settings` modules are now in the toplevel `rdwatch` app
- File structure/other modules are rearranged in accordance with [Resonant best practices](https://github.com/girder/cookiecutter-girder-4)
- The entire app is consolidated into a single Python module, `rdwatch`, with each former app being a submodule of `rdwatch`.

Old module structure:
```
rdwatch/
rdwatch_scoring/
rdwatch_smartflow/
```

New module structure:
```
rdwatch/
	core/
    scoring/
    smartflow/
```

Notable exceptions missing from this PR:
- The default linter/formatter for Resonant is `ruff`; I left our current `flake8`/`black`/`isort` set up alone for now, but we could move to `ruff` in the future. I didn't want to bloat this PR anymore than it already is.
- The default python packaging solution for Resonant is `setuptools`. I don't have a strong opinion on this so I left it as `poetry`, we can discuss later if needed but I'm fine with leaving this part alone as it's worked well for us so far.


rdwatch/urls.py
```diff
@@ -4,16 +4,16 @@
...
 
 # Conditionally add the scoring URLs if the scoring app is installed
-if 'rdwatch_scoring' in apps.app_configs.keys():
-    urlpatterns.append(path('api/scoring/', include('rdwatch_scoring.urls')))
-if 'rdwatch_smartflow' in apps.app_configs.keys():
-    urlpatterns.append(path('api/smartflow/', include('rdwatch_smartflow.urls')))
+if 'rdwatch.scoring' in apps.app_configs.keys():
+    urlpatterns.append(path('api/scoring/', include('rdwatch.scoring.urls')))
+if 'rdwatch.smartflow' in apps.app_configs.keys():
+    urlpatterns.append(path('api/smartflow/', include('rdwatch.smartflow.urls')))
```
BryonLewis, line 16: This isn't working properly.  A dict of the keys in `app.app_configs.keys()` returns the following:
```
dict_keys(['admin', 'auth', 'contenttypes', 'sessions', 'messages', 'staticfiles', 'gis', 'postgres', 'django_extensions', 'django_celery_results', 'allauth', 'account', 'socialaccount', 'gitlab', 'django_rename_app', 'core', 'scoring', 'smartflow', 'debug_toolbar'])
```
Each of the AppConfigs in there has a 'name' property that would match the 'rdwatch.scoring' or the 'rdwatch.smartflow'.  So either we need a way to set the 'key' or instead we search for the name in the app_configs instead of using the key.  There may be a function of `apps` that may make this easier as well.

BryonLewis (changes requested):
Scoring/smartflow endpoints aren't being added.

docker/nginx.json
```diff
@@ -56,9 +56,9 @@
       "type": "python 3.11",
       "user": "rdwatch",
       "group": "rdwatch",
-      "path": "/app/django/",
+      "path": "/app/rdwatch/",
```
mvandenburgh, line 59: Maybe not needed

BryonLewis (approved):

PR #439, opened 2024-06-04
renaming some docker paths to allow local docker compose dev

BryonLewis:


mvandenburgh (approved):
:facepalm: I always forget to update this

PR #443, opened 2024-06-07
Modes update

BryonLewis:
- Updates Modes to be named "Analyst" and "Annotator"
- Uses local storage to store the DB Source Information, either RD-WATCH or Scoring to indicate what datasource is being used
- Creates new Component ModeSelector.Vue to be used to change modes
- The settings panel added the option to change the data source.
- Defaults to using the RD-WATCH datasource.

image


BryonLewis:
TASK:
Disable + Add Proposal in Scoring Proposal view.

vue/src/components/LayerSelection.vue
```diff
@@ -547,7 +547,7 @@ const addProposal = () => {
       :color="state.filters.addingSitePolygon ? 'primary' : ''"
       @click="addProposal()"
     >
-      <v-icon>mdi-plus</v-icon>Add Proposal
+      <v-icon>mdi-plus</v-icon>Add Proposal {{ scoringApp }}
```
mvandenburgh, line 550: Debug leftover?
```suggestion
      <v-icon>mdi-plus</v-icon>Add Proposal
```

mvandenburgh (changes requested):
LGTM just one removal I assume from debugging

vue/src/components/LayerSelection.vue
BryonLewis: Exactly, it was something from yesterday before the demo to remove the Add Proposal from the scoring view.  I've updated it.

mvandenburgh (approved):

PR #446, opened 2024-06-10
User regions

### LINKED ISSUE #445: Region Creation
For SMART runs we decided to utilize the current Region as a limiting factor to allow the user to create custom regions and then run analysis on those Regions.

There are built-in system regions that all have the format of XX_X000.  WE need to make it so you can crate your own regions that have a custom name.

Background/Notes:
- These regions should have a custom name that can't override any of the existing regions.  It should also be relatively short
- Regions will now be modified to allow for a custom name
- I think the Region filter should have a tool for custom filtering RegionIds that are not System based.
- These new Regions should probably also have an 'owner' or a concept of one.  It may not be used initially but I think it should still have it.
-

Process:
- Create/Modify Region model to include the concept of an owner created model.  By default I think we will make all of them public but in the future we can allow users during creation to determine if it is a public or private Region.
- Create endpoint to POST Regions with an owner.
- Develop User Interface and Tool to generate Regions
- Update the Region List to include UserRegions.  Client should be able to show System Regions or User Regions
- Add some test data that utilizes the new regions and posts a model.

BryonLewis:
resolves #445 

- Adds `owner, public` fields to a `region` and `model_run`. Defaults public to `True` and owner to `None` in both instances
- Condenses the `POST` endpoint for Region creation to create a region not associated with model run and being bound to a user with the option to make it Public
- Adds a `GET /region/details` endpoint to get information about owners and public to the list.  This also includes some information about why a user can't delete a region.  They are only able to delete regions they own that have all model runs removed from them.
- Create Region Client Side code.  I think I could modify the existing Region button to have an option for 'Add Region' that allows a user to specify a region with name and the public setting.
- User regions are filters near the top of the region list
- Added a `/regions/{region_id}/vector-tile/z/x/y.pbf` endpoint.  This is to support visualizing regions when no model-runs are associated with them.  This is required to be able to see a region and eventually be able to delete it.
- Drawing of User Regions has been modified to utilize the new /regions/vector-tile endpoint.  I modified region naming so that user regions are indicated by {name}_{userId}.  This is for the query selection as well as making sure regions are unique.  There are new unique constraints so that the combination of name and owner has to be unique.  So now there is a 'value' that is used to indicate a region that is a combination of the name and owner.  If the owner is None (it is a system region uploaded by the admin) then it is just the name.
- I felt it was better to reference the vector-tiles by the region Id instead of the region name.  It is still necessary to use the region name for the URL parameters to make it easy to share regions.  To facilitate this I added a `regionMap` to the store to align a region.
- Redid the Layer Selection of Regions to allow regions to be displayed only if they have geometry associated with them.  It also only allows deletion if the user is the owner and no models are linked to the region.  If you can't delete it there is a tooltip displaying the reason.
- I updated model-run endpoint to provide a boundingbox of the region if no model-runs are connected to a region.  Previously it would just return a null boundingbox
- Connected up deletion of region from the interface
- Moved Region updating to the Store as well as storing of the region list because multiple components need to trigger an update as well as use the values.  RegionFilter no longer is the only thing that will cause an update to the region list.
- Added in basic downloading of a region geojson, right now it sets default values for the custom fields in the region geometry.


BryonLewis:
New Request:
Add the ability to export and download the Region Polygon in geoJSON format.

rdwatch/core/views/model_run.py
```diff
@@ -292,16 +293,15 @@ def paginate_queryset(
...
         model_runs = None
         cache_key = _get_model_runs_cache_key(filters.dict() | pagination.dict())
-        if (
-            'proposal' not in filters.dict().keys()
-        ):  # adjudicated status can't be cached for proposals
-            model_runs = cache.get(cache_key)
+        # if (
+        #     'proposal' not in filters.dict().keys()
+        # ):  # adjudicated status can't be cached for proposals
+        #     model_runs = cache.get(cache_key)
```
mvandenburgh, line 299:
```suggestion
```

rdwatch/core/views/model_run.py
```diff
@@ -315,6 +315,18 @@ def paginate_queryset(
...
+                if qs['count'] == 0:  # No model runs we set bbox to Region bbox
+                    region_filter = filters.dict()['region']
+                    regions = [
+                        obj
+                        for obj in Region.objects.all()
+                        if obj.value == region_filter
+                    ]
+                    if len(regions) > 0:
+                        bbox = regions[0].geojson
+                        model_runs['bbox'] = bbox
```
mvandenburgh, line 328: We can do this in pure SQL instead of python:
```suggestion
                    regions = Region.objects.filter(value=region_filter)
                    if regions.exists():
                        bbox = regions[0].geojson
                        model_runs['bbox'] = bbox
```

rdwatch/core/views/region.py
```diff
@@ -1,14 +1,174 @@
...
+            'name': region[0],
+            'owner': region[1] if region[1] is not None else 'None',
+            'value': region[0] if region[1] is None else f'{region[0]}_{region[1]}',
+            'public': region[2],
+            'id': region[3],
+            'deleteBlock': get_delete_block(region[3], request.user),
+            'hasGeom': True if region[4] else False,
+        }
+        for region in regions
+    ]
```
mvandenburgh, line 56: I think this should use a ninja `Schema` instead of an untyped dict. As it is, it's difficult to parse what's going on with the indexes IMO, and we're also iterating over a queryset when we don't need to.  i.e. changing to something like

```suggestion
@router.get('/details', response=list[RegionDetailResponseSchema])
@paginate(PageNumberPagination, page_size=1000)
def list_region_details(request: HttpRequest):
    regions = Region.objects.values_list(
        'name', 'owner__username', 'public', 'pk', 'geom'
    )
    return regions
```

Where `RegionDetailResponseSchema` is defined as needed.

rdwatch/core/views/region.py
```diff
@@ -1,14 +1,174 @@
...
+    region_model: RegionModel,
+):
+    owner = request.user
+    # Extract the owner from the request user
+    region, _created = Region.create_region_model_from_geoJSON(
+        region_model, True, owner
+    )
+
+    return Response(
+        {'detail': 'Region model created successfully', 'region_id': str(region.name)},
```
mvandenburgh, line 71: `region.name` is not nullable, so it will always be of type `str`.
```suggestion
        {'detail': 'Region model created successfully', 'region_id': region.name},
```

rdwatch/core/views/region.py
```diff
@@ -1,14 +1,174 @@
...
+        status=201,
+    )
```
mvandenburgh, line 73: It should be possible to simplify this to
```suggestion
    return 201, {'detail': 'Region model created successfully', 'region_id': str(region.name)}
```
If we want to return structured output that has a `detail` and `region_id` field, I think this should be defined as a `Schema` instead of using a raw `Reponse` object.

rdwatch/core/views/region.py
```diff
@@ -1,14 +1,174 @@
...
+        return JsonResponse(
+            {
+                'error': 'This region is currently being\
+                used in model runs and cannot be deleted.'
+            },
+            status=400,
+        )
+    # return a successful response for deletion of the region
+    selected_region.delete()
+    return JsonResponse({'success': 'Region deleted successfully.'}, status=200)
```
mvandenburgh, line 97: See previous comment about simplifying the response.

rdwatch/core/views/region.py
```diff
@@ -1,14 +1,174 @@
...
+    )
+
+
+@router.delete('/{region_id}')
+def delete_region(request: HttpRequest, region_id: int):
+    owner = request.user
+    selected_region = get_object_or_404(Region, pk=region_id)
+    if selected_region.owner != owner:  # 403 response, only owner can delete a region
+        return JsonResponse(
+            {'error': 'You do not have permission to delete this region.'}, status=403
```
mvandenburgh, line 82: See previous comment about simplifying the response.

rdwatch/core/views/region.py
```diff
@@ -1,14 +1,174 @@
...
+        region_model, True, owner
+    )
+
+    return Response(
+        {'detail': 'Region model created successfully', 'region_id': str(region.name)},
+        status=201,
+    )
+
+
+@router.delete('/{region_id}')
```
mvandenburgh, line 76:
```suggestion
@router.delete('/{region_id}/')
```

rdwatch/core/views/region.py
```diff
@@ -1,14 +1,174 @@
...
+    return False
+
+
 @router.get('/', response=list[str])
 @paginate(PageNumberPagination, page_size=1000)
 def list_regions(request: HttpRequest):
     return Region.objects.values_list('name', flat=True)
+
+
+@router.get('/details', response=list[dict])
```
mvandenburgh, line 39:
```suggestion
@router.get('/details/', response=list[dict])
```

mvandenburgh (changes requested):
Some comments on the python code from an initial look over, I'll test this locally and do a deeper review that includes the frontend changes next.

rdwatch/core/views/model_run.py
BryonLewis: Ahh I merged this without thinking, value is a property that I generate on the Region Model.  We needed some sort of unique value for Regions as a combination of the Name and the owner.  I tried the same thing initially.  I.E You and I can both have a Region called 'Custom' but we can't have two regions called 'Custom'.  So `value` is a simple concating of `${name}_${owner username} that way it works with our current dropdowns and the URL parameters.

We could swap all of our references to utilize Region Ids (ints) instead of the name.  It just means some more work on the front-end because I would like to keep the name as a URL query parameter like 'rgd.kitware.com/#/KR_R001`.  So I would have to utilize a name to ID mapping on the front end for the request.  Let me know if you want me to do that in this PR or if you need more clarification.

mvandenburgh: We can postpone this for the future, I didn't realize at the time that `value` was a property.

mvandenburgh (approved):

PR #450, opened 2024-07-01
Rdwatch point support

BryonLewis:
Initial enabling of point support, may require some discussion at some point.
This is preparation of having point support in the RDWATCH database as well as enabling the ability to download images around points.

Done:

- Adds a PointField to `SiteEvaluations` and `SiteObservations`
- Allows the `geom` field in `SiteEvaluations, SiteObservations` to be null.  If it is null then it has a `point` field that can be used
- Updated the bulkcreate and the Schemas when posting sites to allow for point based GeoJSON instead of only Polygon or MultiPolygon fields.
- Adds queries for points for `sites` and for `observations`
- Updates the `rdwatchtiles.ts` to add sources for points for both the `sites` and the `observations`
- Creates a new function in `rdwatchtiles.ts` to handle radius sizing based on hover state, groundtruth and selection.  Might want to make a base level that is increased by the user.
- Added to the `mouseEvents.ts` interaction handlers for hovering and clicking both `sites` and `observations`
- When clicking on a site it goes to the `/observations` endpoint to get more details.  I've updated this endpoint to handle the case of having either `geom` or `point`.  Right now if it is a `point` the bounding box is essentially zero.  We may want to handle that on the client side an eventually have a way to zoom to bounding boxes.
- The above was in preparation for eventually handing the downloading of images for the point annotations.
- Changed SiteEvaluation foreign key to ModelRuns to be `CASCADE` instead of `PROTECT`.  I think we may want to make it a bit easier in the future to delete model runs and not have to delete all SiteEvaluations inside of them.

image


rdwatch/core/models/site_evaluation.py
```diff
@@ -44,6 +44,13 @@ class SiteEvaluation(models.Model):
         help_text="Polygon from this site's Site Feature",
         srid=3857,
         spatial_index=True,
+        null=True,
+    )
+    point = PointField(
+        help_text="Point from this site's Site Feature",
+        srid=3857,
+        spatial_index=True,
+        null=True,
```
mvandenburgh, line 53: I think we should add a DB constraint to enforce that one of `geom` or `point` is non-null

mvandenburgh, line 53: I think we should add a DB constraint to enforce that one of `geom` or `point` is non-null

rdwatch/core/models/site_observation.py
```diff
@@ -96,25 +103,46 @@ def bulk_create_from_site_evaluation(
...
+                            label=label_map[phase],
+                            score=feature.properties.score,
+                            geom=polygon,
+                            constellation=constellation,
+                            spectrum=None,
+                            timestamp=feature.properties.observation_date,
+                        )
+                    )
+            if point:
+                point = feature.parsed_geometry
```
mvandenburgh, line 135: Couple things here -

1) `if not point` is more pythonic than `if point is False` 
2) I think this reads cleaner if the negative case comes after. To make it easier in the github editor I just swapped out the boolean to represent whether it's a polygon instead of a point, but either way is good i think
3) Isn't the `point` variable uneeded here (which makes the 1st point here moot)? I believe you can just check `isinstance` directly and get the same outcome.

```suggestion
            is_polygon = False
            if isinstance(feature.parsed_geometry, Polygon | MultiPolygon):
                geometry = (
                    feature.parsed_geometry
                    if isinstance(feature.parsed_geometry, MultiPolygon)
                    else MultiPolygon([feature.parsed_geometry])
                )
                for i, polygon in enumerate(geometry):
                    if feature.properties.current_phase:
                        phase = feature.properties.current_phase[i]
                    else:
                        phase = 'Unknown'

                    site_observations.append(
                        cls(
                            siteeval=site_eval,
                            label=label_map[phase],
                            score=feature.properties.score,
                            geom=polygon,
                            constellation=constellation,
                            spectrum=None,
                            timestamp=feature.properties.observation_date,
                        )
                    )
            else:
                point = feature.parsed_geometry
```

rdwatch/core/models/site_observation.py
```diff
@@ -96,25 +103,46 @@ def bulk_create_from_site_evaluation(
...
                 if feature.properties.current_phase:
-                    phase = feature.properties.current_phase[i]
+                    phase = feature.properties.current_phase[0]
```
mvandenburgh, line 137: Why this change?

rdwatch/core/schemas/site_model.py
```diff
@@ -191,11 +191,11 @@ def parse_geometry(cls, v: dict[str, Any]):
     def ensure_correct_geometry_type(cls, values: dict[str, Any]):
         if 'properties' not in values or 'geometry' not in values:
             return values
-        if (
-            isinstance(values['properties'], SiteFeature)
-            and values['geometry'].get('type') != 'Polygon'
+        if isinstance(values['properties'], SiteFeature) and (
+            values['geometry'].get('type') != 'Polygon'
+            and values['geometry'].get('type') != 'Point'
```
mvandenburgh, line 196:
```suggestion
            values['geometry'].get('type') not in ['Polygon', 'Point']
```

tileserver/rgd.mjs
```diff
@@ -129,6 +129,122 @@ function query(z, x, y, modelRunId, year) {
           ELSE FALSE
```
mvandenburgh, line 129: Need a rebase to eliminate this file

mvandenburgh (changes requested):

rdwatch/core/models/site_observation.py
BryonLewis: Sorry, it gets a bit confusing.

So there is a function that is used to interpret a string delimited list (by ',') of the current phase from the geoJSON.  I didn 't write this but I assume there is a reason for it:
https://github.com/ResonantGeoData/RD-WATCH/blob/rdwatch-point-support/rdwatch/core/schemas/site_model.py#L123-L130https://github.com/ResonantGeoData/RD-WATCH/blob/rdwatch-point-support/rdwatch/core/schemas/site_model.py#L123-L130

In the geom/polygon version, it would loop across the geometry (given that it could be a multi polygon)t and it looks like the code just took the phase in the string-separated list for the last value.  It's weird and I don't know the exact reason for this with multi polygons.  Maybe it was a way to support multiple phase activities without changing DB models in the past?

  For Points there shouldn't be multiple geometry so I just took the first value in the list.

rdwatch/core/models/site_evaluation.py
BryonLewis: :+1: will add to both the site_evaluation and site_observation models.

rdwatch/core/models/site_evaluation.py
```diff
@@ -215,6 +229,14 @@ class Meta:
         default_related_name = 'evaluations'
         indexes = [GistIndex(fields=['timestamp']), GistIndex(fields=['score'])]
 
+        constraints = [
+            CheckConstraint(
+                check=Q(geom__isnull=False) | Q(point__isnull=False),
+                name='site_geom_or_point_not_null',
+                violation_error_message='geom and point cannot both be null',
+            ),
+        ]
```
mvandenburgh, line 238:
```suggestion
        constraints = [
            CheckConstraint(
                check=Q(geom__isnull=False, point__isnull=True) | Q(point__isnull=False, geom__isnull=True),
                name='site_geom_or_point_not_null',
                violation_error_message='Exactly one of [geom, point] must be non-null',
            ),
        ]
```

rdwatch/core/models/site_observation.py
```diff
@@ -127,6 +152,14 @@ class Meta:
         default_related_name = 'observations'
         indexes = [GistIndex(fields=['timestamp']), GistIndex(fields=['score'])]
 
+        constraints = [
+            CheckConstraint(
+                check=Q(geom__isnull=False) | Q(point__isnull=False),
+                name='siteobs_geom_or_point_not_null',
+                violation_error_message='geom and point cannot both be null',
+            ),
+        ]
```
mvandenburgh, line 161:
```suggestion
        constraints = [
            CheckConstraint(
                check=Q(geom__isnull=False, point__isnull=True) | Q(point__isnull=False, geom__isnull=True),
                name='siteobs_geom_or_point_not_null',
                violation_error_message='Exactly one of [geom, point] must be non-null',
            ),
        ]
```

mvandenburgh (changes requested):
Just one suggestion to strengthen the constraints on `point`/`geom` a bit, otherwise this LGTM

mvandenburgh (approved):

PR #456, opened 2024-07-23
Fix downloading UI

### LINKED ISSUE #431: Proposal UI Issues
- [x] When running a new download it should automatically make the currently selected model-run update to have the cancelling option
- [x] The cancellation option in the model-run when clicked should auto dismiss itself once the task is cancelled
- [x] In the Scoring/Proposal view it doesn't properly render the downloading state of the sites after launching a download of images.  It doesn't seem to show the downloading state while it is downloading either.

BryonLewis:
resolves #431 

- Prevents the deselection of items when downloading images for a selected site in scoring view
- Fixes ordering so it properly will order selected to the top after each selection or update
- Adds a global identifier for downloading to indicate to the ModelRun that images are being downloaded so they can be cancelled.


rdwatch/scoring/views/model_run.py
```diff
@@ -507,11 +507,9 @@ def list_model_runs(
 @router.get('/{id}/', response={200: ModelRunDetailSchema})
 def get_model_run(request: HttpRequest, id: UUID4):
     try:
-        EvaluationRun.objects.get(pk=id)
-        data = get_queryset()
+        data = get_object_or_404(get_queryset(), id=id)
```
mvandenburgh, line 510: I don't think `get_object_or_404` raises `EvaluationRun.DoesNotExist` here, so wouldn't this fail prematurely?

BryonLewis: Thanks for finding that, you're correct.  This won't properly swap to the proposal query.

BryonLewis: Decided to give a more descriptive error with missing EvaluationRun/AnnotationProposalSet.

mvandenburgh (approved):

PR #457, opened 2024-07-23
addRegion to Annotator view, fix scoring toggling between modes

### LINKED ISSUE #454: Decide to add or disable Region Additions from Annotator view.

BryonLewis:
resolves #454

- Makes sure the AddRegion side panel comes up in Annotator Mode
- Fixes some toggling between Analyst and Annotator Modes while using the scoring database.


vue/src/components/ModeSelector.vue
```diff
@@ -16,15 +16,15 @@ const proposalView = computed(() => route.path.includes('proposals'));
...
+      :to="scoringApp ? '/scoring/' : '/'"
       :color="!proposalView? 'primary': ''"
       :theme="!proposalView? 'dark': ''"
       class="mx-2"
     >
       Analyst
     </v-btn>
     <v-btn
-      to="/proposals"
+      :to="scoringApp ? '/scoring/proposals' : '/proposals'"
```
mvandenburgh, line 27: Isn't this just re-implementing `ApiService.getApiPrefix()`?
```suggestion
      :to="`${ApiService.getApiPrefix()}/proposals`"
```

mvandenburgh (changes requested):

vue/src/components/ModeSelector.vue
BryonLewis: Yes, I think it's a habit from my other projects.  I utilize `defineComponent` instead of `<script setup>` that way I can be explicit with the `setup()` functions return values.  It just doesn't feel right to me to provide a whole Static Class like `ApiService` to the template so I typically change it to have a property I'm looking for in a simple computed value and return that.

There's also the fact that I'm not importing or using `ApiService` at all in this component.  I'm utilizing the `route` because we don't use `ApiService.getPrefix()` with `/proposals`.

mvandenburgh (approved):

PR #458, opened 2024-07-24
Build docs in CI

mvandenburgh:
This will cause CI to fail if the docs build fails.


mvandenburgh:
Making this a seperate PR just to make sure it works first. @BryonLewis feel free to merge into your branch if you agree with this

BryonLewis (approved):

PR #460, opened 2024-07-29
Properly serve static files in production with `whitenoise`

mvandenburgh:
This will fix the CSS issues on https://rgd.kitware.watch/admin/.

This PR is largely based on https://github.com/kitware-resonant/django-composed-configuration/blob/master/composed_configuration/_static.py

### DIFF
--- a/docker/nginx.json
+++ b/docker/nginx.json
@@ -46,6 +46,11 @@
       }
     },
     {
+      "match": {
+        "uri": "/staticfiles/*"
+      }
+    },
+    {
       "action": {
         "share": "/app/vue/dist/$uri"
       }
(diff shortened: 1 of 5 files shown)


docker/nginx.json
```diff
@@ -45,6 +45,11 @@
         "pass": "applications/rdwatch"
       }
     },
+    {
+      "match": {
+        "uri": "/staticfiles/*"
```
BryonLewis, line 50: should this be /static/  The `STATIC_URL` in the settings.py is `static` the `STATIC_ROOT` which I think is the file location is `staticfiles`?

BryonLewis (changes requested):
Just a question on the nginx route matching

mvandenburgh:
This isn't working within docker, marking as draft until I can look into this

mvandenburgh:
Looking back into this, I don't think we even need `whitenoise` here. That package is used to serve static files from Python directly, and is useful for when you're running in an environment such as Heroku. But we're already using nginx, so we can just use that instead. I'll open another PR with that.

PR #463, opened 2024-07-30
Fix docker configuration for ARM CPUs

zachmullen:
In particular, this makes the full docker-compose dev env build correctly on M1 Macs.

### DIFF
--- a/docker-compose.override.yaml
+++ b/docker-compose.override.yaml
@@ -68,6 +68,7 @@ services:
       django:
         condition: service_healthy
   celery_beat:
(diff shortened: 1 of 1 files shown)


BryonLewis (approved):
- Pulled on a M3 Mac.
- Made sure my `DOCKER_DEFAULT_PLATFORM` environment variable was cleared
- Did a `docker compose build --no-cache` and the build completed successfully.
- Confirmed on the same system that `main` has an error regarding torch+cpu and the `celery_beat` service.

PR #468, opened 2024-08-06
Server Animation Export

BryonLewis:
Eventual PR to support exporting Animated Gifs for sites.

Process:
- Get single export working with basic drawing of polygons on images
- Coloring polygon based on activity label
- Add text information (Satellite Source, Time, Obs vs Non Obs)
- Get filtering working (Satellite Source, Cloud Cover, Obs vs Non Obs)
- Rescaling to center polygon between WV/S2 and different image sizes
- Add in options (Rescaling, Speed, Quality)
- Generate Model-Run level exporting 
    - Can you annotationExport model for storing the zip until downloaded
- Client Side UI for downloading
    - Component to configure settings (filters, speed, quality, rescaling)
    - Same component can be used for single site or for model-run downloading
    - May want to configure a combination of downloading images and creating gifs.

Done:

- Basic testing of single export of the cropped images.  It creates a GIF of the images but doesn't properly draw polygons or rescale the images.
- Draws polygons in basic the scale without the proper coloring for the label
- Added simple date text drawing
- Rescaling of Images for S2/WV different sizes
- Rescaling Border for shrinking sizes
- Handling text display with automatic font size to different resolutions
- Options to toggle on/off different labels

TODO:

- [x] Polygon/Geometry Rescaling
- [x] Additional Filters (cloudcover, NoData, obs vs nonobs, activity labels)
- [x] Basic dialog for exporting animation from the server
- [x] Create model-run task to export multiple sites in parallel into a zip file for downloading

Notes About Rescaling:

The **Rescaling** option will allow the system to export all of the images while centered with the same spatial view on the bounding box.  This means that WV/S2/L8 images of different resolutions can all be used but the polygon remains a similar size on the screen.

This is done by taking the highest-resolution image in the list and using it as a basis.  All S2/L8 or other low-resolution images are rescaled so that their polygons and pixels match the highest-resolution image.   The polygon/geom is drawn after and not before to make sure it doesn't history too much.

There is a **Rescale Border** option that will add additional geospatial padding around the highest resolution image so that lower-resolution images have more context.  This requires that the highest resolution image add blank NoData or gray data around it and that the lower resolution images now have more padding.

Rescaling Process:

1. Using the filters for sources, cloud clover, nodata and any other items find the image with the highest resolution - HR.
2. HR has some space around the geom (point/polygon) we have a mapping of the image bounds to geospatial bounds (geobbox).
3. If we have a **Rescale Border** we increase the spatial area around the image by multiplying it by the geobbox and we get a BaseResolution.  I.E a Rescale Border values of 2 makes the area 2x the existing image bbox.
4. All images generate are now the BaseResolution.  Images are cropped and draw into the BaseResolution image and need to be properly centered.
5. Go through the existing satellite images, crop them based on the geospatial bounds and place them inside the Base Resolution in the appropriate location.


rdwatch/core/models/annotation_exports.py
```diff
@@ -24,8 +22,44 @@ class AnnotationExport(models.Model):
...
+
+class AnimationSiteExport(AbstractFileExport):
+    site_evaluation = models.ForeignKey(
+        to='SiteEvaluation',
+        on_delete=models.PROTECT,
+        help_text='SiteEvaluation for Animation export',
+        db_index=True,
+    )
+    arguments = models.JSONField(null=True, default=None)
+    user = models.ForeignKey(User, on_delete=models.CASCADE, null=True, blank=True)
```
mvandenburgh, line 54: Does this need to allow `null` values? I think there will always be an associated user, right?
```suggestion
    user = models.ForeignKey(User, on_delete=models.CASCADE)
```

rdwatch/core/models/annotation_exports.py
```diff
@@ -24,8 +22,44 @@ class AnnotationExport(models.Model):
...
+
+
+class AnimationModelRunExport(AbstractFileExport):
+    configuration = models.ForeignKey(
+        to='ModelRun',
+        on_delete=models.PROTECT,
+        help_text='ModelRun for Animation export',
+        db_index=True,
+    )
+    user = models.ForeignKey(User, on_delete=models.CASCADE, null=True, blank=True)
```
mvandenburgh, line 64: Same question as above
```suggestion
    user = models.ForeignKey(User, on_delete=models.CASCADE)
```

rdwatch/core/tasks/__init__.py
```diff
@@ -461,6 +467,14 @@ def collect_garbage_task() -> None:
...
     ).delete()
 
+    AnimationSiteExport.objects.filter(
+        created__lte=timezone.now() - timedelta(hours=6)
+    ).delete()
+
+    AnimationModelRunExport.objects.filter(
+        created__lte=timezone.now() - timedelta(hours=48)
+    ).delete()
+
```
mvandenburgh, line 477: What do you think about adding some info in the UI indicating when these will "expire"?

rdwatch/core/views/animation.py
```diff
@@ -0,0 +1,218 @@
...
+)
+
+logger = logging.getLogger(__name__)
+
+router = Router()
+
+
+def generate_download_data(
+    exports: list[AnimationSiteExport] | list[AnimationModelRunExport],
+):
```
mvandenburgh, line 25: These type annotations are incorrect. To be exact, you're passing `QuerySets` as `exports` where this function is used -
```suggestion
from django.db.models import QuerySet

def generate_download_data(
    exports: QuerySet[AnimationSiteExport] | QuerySet[AnimationModelRunExport],
):
```

Alternatively, this can be typed more abstractly -
```suggestion
from typing import Iterable

def generate_download_data(
    exports: Iterable[AnimationSiteExport] | Iterable[AnimationModelRunExport],
):
```

rdwatch/core/views/animation.py
```diff
@@ -0,0 +1,218 @@
...
+def delete_site_download(
+    request: HttpRequest,
+    id: UUID4,
+):
+    # Fetch the export object
+    export = get_object_or_404(AnimationSiteExport, celery_id=id, user=request.user)
+    logger.warning(export)
+    # Check for ownership or permissions
+    if export.user != request.user:
+        return 403, {'error': 'You do not have permission to delete this export.'}
```
mvandenburgh, line 117: Since you're including `user` in the `get_object_or_404` query, the `export` will always belong to this user, or a 404 will be returned. 

I'm not sure if that was intentional or not, but I think that behavior is fine. I actually prefer this pattern over returning a 403, since returning a 403 to an unauthorized user is actually leaking the existence of the object to them. not that it matters much in this specific case, but in general I usually go with this strategy.

fwiw, github does the same thing; notice it will return a 404 if you try to access a private repo while logged out.

```suggestion
    # Fetch the export object
    export = get_object_or_404(AnimationSiteExport, celery_id=id, user=request.user)
```

rdwatch/core/views/animation.py
```diff
@@ -0,0 +1,218 @@
...
+    datapath, name = create_animation(
+        site_evaluation_id=id,
+        settings=params.dict(),
+    )
+    if datapath:
+        with open(datapath, 'rb') as f:
+            response = HttpResponse(f.read(), content_type='application/octet-stream')
+            response['Content-Disposition'] = f'attachment; filename={name}'
+            return response
+    return 500, 'Unable to export data'
```
mvandenburgh, line 159:
```suggestion
```

rdwatch/core/views/animation.py
```diff
@@ -0,0 +1,218 @@
...
+        animation_export = animation_export.first()
+        name = animation_export.name
+        content_type = 'video/mp4'
+        if name.endswith('.gif'):
+            content_type = 'image/gif'
+        response = HttpResponse(
+            animation_export.export_file.file, content_type=content_type
+        )
+        response['Content-Disposition'] = f'attachment; filename="{name}"'
+        return response
```
mvandenburgh, line 190:
```suggestion
    animation_export = get_object_or_404(AnimationSiteExport, celery_id=task_id, user=request.user)
    name = animation_export.name
    content_type = 'video/mp4'
    if name.endswith('.gif'):
        content_type = 'image/gif'
    response = HttpResponse(
        animation_export.export_file.file, content_type=content_type
    )
    response['Content-Disposition'] = f'attachment; filename="{name}"'
    return response
```

rdwatch/core/views/animation.py
```diff
@@ -0,0 +1,218 @@
...
+        celery_id=task_id, user=request.user
+    )
+    if animation_export.exists():
+        animation_export = animation_export.first()
+        name = animation_export.configuration.title
+        response = HttpResponse(
+            animation_export.export_file.file, content_type='application/zip'
+        )
+        response['Content-Disposition'] = f'attachment; filename="{name}.zip"'
+        return response
```
mvandenburgh, line 205:
```suggestion
    animation_export = get_object_or_404(AnimationModelRunExport, celery_id=task_id, user=request.user)
    name = animation_export.configuration.title
    response = HttpResponse(
        animation_export.export_file.file, content_type='application/zip'
    )
    response['Content-Disposition'] = f'attachment; filename="{name}.zip"'
    return response
```

mvandenburgh (changes requested):
I figured it would be better to post some actionable feedback now instead of holding it back and dropping everything at once, so I did an initial read-through of the code (except for `rdwatch/core/tasks/animation_export.py`) and left some initial review comments/suggestions. Next, I'll review the animation_export task and do some testing locally and leave any further feedback.

rdwatch/core/tasks/__init__.py
BryonLewis: Yup.  Thanks I forgot to even mention that in the videos.  It could be added to the download data page (simple calculation of the created time and the now time).

rdwatch/core/views/animation.py
BryonLewis: I'll remove this before we merge.  Once we have images again this is a quick endpoint to utilize if you need to test any changes to the animation drawing and don't want to deal with celery container reloading.

rdwatch/core/views/animation.py
BryonLewis: Yup at some point I think I did convert them into a list, but removed it.

rdwatch/core/tasks/__init__.py
BryonLewis: I added an expired date to the Download listing as well as a brief sentence indicating downloads will be removed (6 hours for individual sites, 48 hours for modelruns).

rdwatch/core/views/animation.py
```diff
@@ -0,0 +1,211 @@
...
+def get_downloaded_modelrun_animation(request: HttpRequest, task_id: UUID4):
+    animation_export = get_object_or_404(
+        AnimationModelRunExport, celery_id=task_id, user=request.user
+    )
+    name = animation_export.configuration.title
+    response = HttpResponse(
+        animation_export.export_file.file, content_type='application/zip'
+    )
+    response['Content-Disposition'] = f'attachment; filename="{name}.zip"'
+    return response
```
mvandenburgh, line 198: I think we should avoid buffering the entire file into memory in a server request thread. Instead, the usual pattern is to generate a presigned url and redirect to s3/minio.

```suggestion
from django.shortcuts import redirect

@router.get('/download/site/{task_id}/')
def get_downloaded_site_animation(request: HttpRequest, task_id: UUID4):
    return redirect(get_object_or_404(
        AnimationSiteExport, celery_id=task_id, user=request.user
    ), permanent=False)


@router.get('/download/modelrun/{task_id}/')
def get_downloaded_modelrun_animation(request: HttpRequest, task_id: UUID4):
    return redirect(get_object_or_404(
        AnimationModelRunExport, celery_id=task_id, user=request.user
    ).export_file.url, permanent=False)
```

@BryonLewis is it necessary to set the Content-Disposition headers? If so, I think we can do that but it makes things a bit more complicated.

rdwatch/core/tasks/animation_export.py
```diff
@@ -0,0 +1,802 @@
...
+    """
+    minx, miny, maxx, maxy = bbox
+
+    center_x = (minx + maxx) / 2
+    center_y = (miny + maxy) / 2
+
+    width = (maxx - minx) * scale / 2
+    height = (maxy - miny) * scale / 2
+
+    return [center_x - width, center_y - height, center_x + width, center_y + height]
```
mvandenburgh, line 131: The `.extent` property from a geodjango polygon is actually a `tuple`, not a `list`. I would also suggest changing the return type to be consistent with that.

```suggestion
def rescale_bbox(bbox: tuple[float, float, float, float], scale: float) -> tuple[float, float, float, float]:
    """
    Rescales a bounding box by a given scaling factor.

    Parameters:
    bbox (tuple): A tuple containing [minx, miny, maxx, maxy].
    scale (float): The scaling factor to apply to the bounding box.

    Returns:
    list: A list containing the rescaled [minx, miny, maxx, maxy].
    """
    minx, miny, maxx, maxy = bbox

    center_x = (minx + maxx) / 2
    center_y = (miny + maxy) / 2

    width = (maxx - minx) * scale / 2
    height = (maxy - miny) * scale / 2

    return (center_x - width, center_y - height, center_x + width, center_y + height)
```

rdwatch/core/tasks/animation_export.py
```diff
@@ -0,0 +1,802 @@
...
+    pixel_per_unit_y (float): Pixels per unit for the Y-axis of the source image.
+    output_width (int): Width of the output image.
+    output_height (int): Height of the output image.
+    output_bbox (list): Bounding box of the output image in the format [minx, miny, maxx, maxy].
+    output_pixel_per_unit_x (float): Pixels per unit for the X-axis of the output image.
+    output_pixel_per_unit_y (float): Pixels per unit for the Y-axis of the output image.
+
+    Returns:
+    Image: The output image with the source image pasted in.
+    """
```
mvandenburgh, line 162:
```suggestion
def paste_image_with_bbox(
    source_image: Image,
    bbox: tuple[float, float, float, float],
    pixel_per_unit_x: float,
    pixel_per_unit_y: float,
    output_width: int,
    output_height: int,
    output_bbox: tuple[float, float, float, float],
    output_pixel_per_unit_x: float,
    output_pixel_per_unit_y: float,
) -> Image:
    """
    Pastes a source image into an output image based on matching bounding box coordinates,
    with different pixel-per-unit scales for the X and Y axes.

    Parameters:
    source_image_path (Image): source image.
    bbox (tuple): Bounding box of the source image in the format [minx, miny, maxx, maxy].
    pixel_per_unit_x (float): Pixels per unit for the X-axis of the source image.
    pixel_per_unit_y (float): Pixels per unit for the Y-axis of the source image.
    output_width (int): Width of the output image.
    output_height (int): Height of the output image.
    output_bbox (list): Bounding box of the output image in the format [minx, miny, maxx, maxy].
    output_pixel_per_unit_x (float): Pixels per unit for the X-axis of the output image.
    output_pixel_per_unit_y (float): Pixels per unit for the Y-axis of the output image.

    Returns:
    Image: The output image with the source image pasted in.
    """
```

rdwatch/core/tasks/animation_export.py
```diff
@@ -0,0 +1,802 @@
...
+    gray_color = (gray_value, gray_value, gray_value, 0)
+    output_image = Image.new('RGBA', (output_width, output_height), gray_color)
+
+    # Paste the source image into the output image
+    output_image.paste(cropped_img)
+
+    return output_image
+
+
+def to_pixel_coords(lon, lat, bbox, xScale, yScale, xOffset=0, yOffset=0):
```
mvandenburgh, line 200:
```suggestion
def to_pixel_coords(lon: float, lat: float, bbox: tuple[float, float, float, float], xScale: float, yScale: float, xOffset=0, yOffset=0):
```

rdwatch/core/tasks/animation_export.py
```diff
@@ -0,0 +1,802 @@
...
+def draw_text_in_box(
+    draw,
+    text,
+    position,
+    box_size,
+    box_color=(80, 80, 80),
+    text_color=(255, 255, 255),
+    font_path=None,
+    initial_font_size=30,
+):
```
mvandenburgh, line 215:
```suggestion
def draw_text_in_box(
    draw: ImageDraw,
    text: str,
    position: tuple[float, float],
    box_size: tuple[float, float],
    box_color=(80, 80, 80),
    text_color=(255, 255, 255),
    font_path: str | None = None,
    initial_font_size=30,
):
```

mvandenburgh (changes requested):
I tested it out and everything works as expected for me. Just added some suggestions for performance improvements, and encoding the parameter/return types specified in the doc strings as actual type annotations.

rdwatch/core/views/animation.py
BryonLewis: I'd have to double check if without attachment it properly downloads.

BryonLewis: Looks like it won't handle it properly.

rdwatch/core/views/animation.py
```diff
@@ -0,0 +1,211 @@
...
+            )
+        else:
+            task_id = export.celery_id
+            task = AsyncResult(task_id)
+            celery_data = {}
+            celery_data['state'] = task.state
+            celery_data['status'] = task.status
+            celery_data['info'] = (
+                str(task.info) if isinstance(task.info, RuntimeError) else task.info
+            )
```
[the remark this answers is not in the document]
BryonLewis: Didn't think this one through and it breaks the logging for the state for the download.  Changing it to `Exception` instead of `runtimeError` should hopefully catch all errors and stringify them.  The task info is an object that contains the current/total and modes.

rdwatch/core/views/animation.py
mvandenburgh: I think we should still fix this, but I'm fine with merging this as-is to avoid bloating this PR more as the solution might not be straightforward.

@zachmullen this is what I was referring to when we were discussing generated presigned urls w/ a Content-Disposition header set. I'm curious what your thoughts are on leaving this as-is vs. generating a presigned url with correct headers (at the cost of more code complexity)? My instinct is that the in-memory approach could cause unforeseen issues with memory usage/response times, but it's also possible I'm overthinking this.

zachmullen: To avoid the issue with memory, use a [`django.http.response.FileResponse`](FileResponse) rather than a normal `HttpResponse`. That will avoid the issue of loading the whole file into memory. The downside of this approach is that it will require Django connections to be occupied serving binary download of potentially large files, which is something we typically want to avoid, particularly under WSGI as opposed to ASGI. From the Django docs:

> There are performance considerations when doing this, though. Django, under WSGI, is designed for short-lived requests. Streaming responses will tie a worker process for the entire duration of the response. This may result in poor performance.

Maybe the best thing to do is have a sync-up to understand the requirements around your content-disposition headers in the first place.

BryonLewis: There may be a way I can directly reference the presigned URL.  I didn't have the time to properly debug it earlier today so I tried the suggested code and swapped back after it didn't work.

BryonLewis: I modified the front-end so I can send it just the pre-signed URL for downloading and a filename to use.
I just need to do a bit more testing on it tomorrow morning,

mvandenburgh: Perfect, looks like this approach works :+1:

mvandenburgh (approved):

PR #470, opened 2024-08-06
Add versioning info

floryst:
- Populate versioning info in the docker build for prod and dev
- Show the version in the UI
  image

The dev build will still show `0.0.0dev0` because `/app/rdwatch/__init__.py` doesn't exist in the docker container at build time (and it gets mounted over anyways). I verified that the prod docker image has a populated `__init__.py`, but haven't tried running it yet. Installing `poetry-dynamic-versioning` and running `poetry dynamic-versioning` will update the version info.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -78,6 +78,7 @@ RUN mkdir /app/rdwatch \
  && touch /app/rdwatch/smartflow/__init__.py \
  && touch /app/README.md \
  && poetry install --only main
+RUN poetry self add poetry-dynamic-versioning@^1.4.0
 
 
 # Build stage that also installs dev dependencies.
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -78,6 +78,7 @@ RUN mkdir /app/rdwatch \
  && touch /app/rdwatch/smartflow/__init__.py \
  && touch /app/README.md \
  && poetry install --only main
+RUN poetry self add poetry-dynamic-versioning@^1.4.0
```
mvandenburgh, line 81: Is this needed? We specify `poetry-dynamic-versioning` in [the PEP 518 `build_system` section of pyproject.toml](https://github.com/ResonantGeoData/RD-WATCH/blob/main/pyproject.toml#L1-L2), is that not installed automatically with `poetry install --only main`?

mvandenburgh: Hmm, I tested it locally and it seems it doesn't install it automatically... so is the idea here that we don't want to include this as a runtime dependency, so we just `poetry add` it dynamically at build time?

floryst: - I can only seem to get poetry to install from `build-system.requires` when I have `[tool.poetry.build]` with an empty build script specified. I don't know how else to get it to install those required packages, and that config key is still an undocumented feature.
- I didn't try to further explore keeping the package out of the final docker image. Right now it's still part of the final image since `poetry self add` still adds the package to the same virtualenv dir as everything else. In that case, it might be nicer to install it as part of deps if you want to generate versions in dev.

mvandenburgh (approved):

PR #472, opened 2024-08-08
Preliminary support for  WorldView NITF imagery

mvandenburgh:
This PR adds support for downloading WorldView NITF imagery. 

On the STAC server, NITFs are stored in the `worldview-nitf` collection. I created a new module, `rdwatch.core.utils.worldview_nitf`, that mirrors the existing `rdwatch.core.utils.worldview_processed` with a few key differences:

1) I removed pan-sharpening logic from `worldview_nitf`.
2) `rio-tiler` gets stuck indefinitely if you give it an S3 URL to a NITF file directly. I assume it's struggling with the size somehow, as obviously unlike COGs it must download the entire image (on the order of gigabytes) every time it wants to chip out a small bounding box. To get around this, I download the NITF image from S3 directly using boto3 to a temporary file and pass `rio-tiler` the path to that, and it works.
3) Following on from 2), because we need to do that explicit image download to the worker's filesystem, it is not really possible to have the region-level on-demand imagery from an endpoint. Thus the only way to use NITF imagery in this PR is in the site/observation-level chips; see the GIF I posted in Slack.

Going forward I'd like to clean up some of the image fetching architecture and logic **(especially if we're going to add dev docs on how this stuff is structured/how to add new image sources)**. Currently there are a lot of complicated functions that are tighly coupled to each other and difficult to reason about. But for the purposes of keeping this PR well-scoped, I haven't made any attempts at that here.

### DIFF
--- a/rdwatch/core/utils/worldview_nitf/satellite_captures.py
+++ b/rdwatch/core/utils/worldview_nitf/satellite_captures.py
@@ -1,0 +1,58 @@
+from dataclasses import dataclass
+from datetime import datetime, timedelta
+from typing import cast
+
+from rdwatch.core.utils.worldview_nitf.stac_search import worldview_search
+
+
+@dataclass()
+class WorldViewNITFCapture:
+    timestamp: datetime
+    bbox: tuple[float, float, float, float]
+    uri: str
+    cloudcover: int | None
+    collection: str
+    bits_per_pixel: int
+    epsg: int
+
+
+def get_features(
+    timestamp: datetime,
+    bbox: tuple[float, float, float, float],
+    timebuffer: timedelta,
+):
+    results = worldview_search(timestamp, bbox, timebuffer=timebuffer)
+    yield from results['features']
+
+
+def get_captures(
+    timestamp: datetime,
+    bbox: tuple[float, float, float, float],
+    timebuffer: timedelta | None = None,
+) -> list[WorldViewNITFCapture]:
+    if timebuffer is None:
+        timebuffer = timedelta(hours=1)
+
+    features = [f for f in get_features(timestamp, bbox, timebuffer=timebuffer)]
+    captures = []
+    for feature in features:
+        if 'data' in feature['assets']:
+            cloudcover = 0
+            if 'properties' in feature:
+                if 'eo:cloud_cover' in feature['properties']:
+                    cloudcover = feature['properties']['eo:cloud_cover']
+            # from pprint import pprint; pprint(feature)
+            capture = WorldViewNITFCapture(
+                timestamp=datetime.fromisoformat(
+                    feature['properties']['datetime'].rstrip('Z')
+                ),
+                bbox=cast(tuple[float, float, float, float], tuple(feature['bbox'])),
+                uri=feature['assets']['data']['href'],
+                bits_per_pixel=feature['properties']['nitf:bits_per_pixel'],
+                cloudcover=cloudcover,
+                collection=feature['collection'],
+                epsg=feature['properties']['proj:epsg'],
+            )
+            captures.append(capture)
+
+    return captures
--- a/rdwatch/core/utils/worldview_nitf/satellite_captures.py
+++ b/rdwatch/core/utils/worldview_nitf/satellite_captures.py
@@ -41,7 +41,6 @@ def get_captures(
             if 'properties' in feature:
                 if 'eo:cloud_cover' in feature['properties']:
                     cloudcover = feature['properties']['eo:cloud_cover']
-            # from pprint import pprint; pprint(feature)
             capture = WorldViewNITFCapture(
                 timestamp=datetime.fromisoformat(
                     feature['properties']['datetime'].rstrip('Z')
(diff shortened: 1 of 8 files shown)


rdwatch/core/utils/worldview_nitf/satellite_captures.py
```diff
@@ -0,0 +1,58 @@
...
+
+    features = [f for f in get_features(timestamp, bbox, timebuffer=timebuffer)]
+    captures = []
+    for feature in features:
+        if 'data' in feature['assets']:
+            cloudcover = 0
+            if 'properties' in feature:
+                if 'eo:cloud_cover' in feature['properties']:
+                    cloudcover = feature['properties']['eo:cloud_cover']
+            # from pprint import pprint; pprint(feature)
```
zachmullen, line 44:
```suggestion
```

mvandenburgh: :grimacing:

vue/src/components/ImagesDownloadDialog.vue
```diff
@@ -19,8 +19,10 @@ const emit = defineEmits<{
     (e: "cancel"): void;
 }>();
 
-const baseList = ref(['S2', 'WV', 'L8', 'PL'])
-const selectedSource: Ref<Constellation[]> = ref(['WV']);
+const constellationChoices = ref(['S2', 'WV', 'L8', 'PL'])
+const selectedConstellation: Ref<Constellation[]> = ref(['WV']);
+const imageSourceChoices = computed<string[] | null>(() => selectedConstellation.value.includes('WV') ? ['cog', 'nitf'] : null);
+const selectedImageSource = ref<'cog' | 'nitf' | undefined>(undefined);
```
BryonLewis, line 25: Without a default value for WV you get an error.  This is related to your `@validator` not working for worldViewSource.

rdwatch/core/views/site_observation.py
```diff
@@ -215,6 +216,14 @@ class GenerateImagesSchema(Schema):
...
     bboxScale: None | float = 1.2
 
+    @validator('worldviewSource')
+    def validate_worldview_source(cls, v: str | None, values: dict[str, Any]):
+        if 'WV' in values['constellation'] and v is None:
+            raise ValidationError('worldviewSource is required for WV constellation')
+        elif 'WV' not in values['constellation'] and v is not None:
+            raise ValidationError('worldviewSource is only for WV constellation')
+        return v
+
```
BryonLewis, line 226: I don't think this validator is working properly.  I'm able to provide a None value for the worldviewSource and it doesn't error out.  It attempts to run it and the task fails because it uses the base search (PL, L8, S2) for the satellite images when in WV mode.

BryonLewis (changes requested):
I pulled and tested and saw the nitfs 'eventually' download and were displayed in the visualization tool.
There is an issue with not selecting a worldViewSource and running the task will cause it to fail the validator and then cause some errors in the task.

vue/src/components/ImagesDownloadDialog.vue
```diff
@@ -82,12 +85,19 @@ const display = ref(true);
...
               label="Source"
               class="mr-2"
             />
+            <v-select
+              v-if="imageSourceChoices"
+              v-model="selectedImageSource"
+              :items="imageSourceChoices"
+              label="Imagery Type"
+              class="mr-2"
+            />
```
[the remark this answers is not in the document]
mvandenburgh: discussed in person, will defer to a few PR

vue/src/components/ImagesDownloadDialog.vue
```diff
@@ -19,8 +19,10 @@ const emit = defineEmits<{
     (e: "cancel"): void;
 }>();
 
-const baseList = ref(['S2', 'WV', 'L8', 'PL'])
-const selectedSource: Ref<Constellation[]> = ref(['WV']);
+const constellationChoices = ref(['S2', 'WV', 'L8', 'PL'])
+const selectedConstellation: Ref<Constellation[]> = ref(['WV']);
+const imageSourceChoices = computed<string[] | null>(() => selectedConstellation.value.includes('WV') ? ['cog', 'nitf'] : null);
```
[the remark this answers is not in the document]
mvandenburgh: https://github.com/ResonantGeoData/RD-WATCH/pull/472/commits/3ba174d17c859a35365108e564da8d50e21fdeb6

rdwatch/core/views/site_observation.py
mvandenburgh: There was two things happening here -
1) This validator wasn't being called at all.
2) If it did get called, accessing `values['constellation'] ` wouldn't have worked because it wouldn't have been populated yet from what i understand reading the pydantic docs. Instead, we need `root_validator` to access multiple fields in one validator.

I'm still not sure why 1) happened, but switching to `root_validator` seems to have fixed this :shrug: https://github.com/ResonantGeoData/RD-WATCH/pull/472/commits/0a9d3489b250e9af5311d407e5b7c9e9cde88c59

vue/src/components/ImagesDownloadDialog.vue
mvandenburgh: Added default https://github.com/ResonantGeoData/RD-WATCH/pull/472/commits/ed700e6f07ad83a2f712ab489d118403e98dbfce

BryonLewis (approved):
Pulled and tested the latest.  The WV source being empty properly will error out now.

PR #473, opened 2024-08-09
Do a full checkout in docker image build

mvandenburgh:
`poetry-dynamic-versioning` requires a full checkout of the repo to determine the version string.


floryst (approved):
Let's try it! We'll revert if it doesn't seem to work.

PR #477, opened 2024-08-12
Update format of version string

mvandenburgh:
This PR updates the version string format used by `poetry-dynamic-versioning` to be `semver` format. Then, in the application code, we parse the string using `python-semver` and render it in this format:
`v{tag}.dev+{distance} ({commit})`

  - **Tag:** The base version tag (e.g., `v1.10`).
  - **Distance:** The number of commits since the last tag, prefixed by `.dev+`. This part is only included if there are additional commits beyond the tag.
  - **Commit Hash:** The first 7 characters of the current commit hash, enclosed in parentheses.
  
**Examples:**
- **Exact Tag:** If the current commit matches the latest tag (`v1.10`), the version string will be:
  ```
  v1.10 (47e928d)
  ```
- **Commits After Tag:** If there are additional commits after the latest tag (e.g., 3 commits after `v1.10`), the version string will be:
  ```
  v1.10.dev+3 (47e928d)
  ```

I'll build the resulting image on my forked repo and deploy it to the dev site to test this, to verify that it works before merging.


mvandenburgh:
Superseded by #504

PR #479, opened 2024-08-12
Remove unused variable from vite.config.ts

zachmullen:
This fixes an eslint warning.


BryonLewis:
I made a small PR into this just to resolve the other warning being displayed during reviews.

BryonLewis (approved):
Thanks for doing this, been meaning to for a while.

For background:  We were at one point using a node.js implementation of the vector server to reduce the load on the django server.  Turns out it wasn't the django server having the issue but a lambda function that was placed in between the user and the server to process requests and make sure that gitlab authentication worked.  Mike updated this to integrate it in the main application and we no longer had a need for the nodeJS vector server.  We just forgot to remove that variable.

PR #482, opened 2024-08-14
Update query for site point dates

JHUAPL-hjg1:


rdwatch/scoring/views/vector_tile.py
```diff
@@ -526,7 +526,7 @@ def vector_tile(
         evaluation_run_uuid, z, x, y, latest_timestamp
     )
 
-    tile = cache.get(cache_key)
+    tile = None
```
mvandenburgh, line 529:
```suggestion
    tile = cache.get(cache_key)
```

Leftover from development?

mvandenburgh (approved):
LGTM, I pulled and tested with a point-based evaluation in my local scoring DB and everything seems to work as intended :+1:

PR #485, opened 2024-08-14
Relax name restrictions on model run region names

### LINKED ISSUE #484: Relax name restrictions on model run region names

zachmullen:
Fixes #484

For now I've just made the region name accept any non-empty string that fits in the database column. Let me know if that's not constrained enough.


BryonLewis (approved):
Now that the other PR is merged, I believe this is good to go.

PR #487, opened 2024-08-14
Update query for model run details

JHUAPL-hjg1:


rdwatch/scoring/views/model_run.py
```diff
@@ -507,9 +510,9 @@ def list_model_runs(
 @router.get('/{id}/', response={200: ModelRunDetailSchema})
 def get_model_run(request: HttpRequest, id: UUID4):
     if EvaluationRun.objects.filter(pk=id).exists():
-        data = get_queryset()
+        data = get_queryset(id=id).first()
     elif AnnotationProposalSet.objects.filter(pk=id).exists():
-        data = get_object_or_404(get_queryset_proposal(), id=id)
+        data = get_object_or_404(get_queryset_proposal(), id=id).first()
```
mvandenburgh, line 515: `get_object_or_404` returns an instance of `AnnotationProposalSet`, not `QuerySet[AnnotationProposalSet]`, so the `.first()` is not needed here and will cause a crash
```suggestion
        data = get_object_or_404(get_queryset_proposal(), id=id)
```

PR #488, opened 2024-08-14
Fix whitespace in string, bump max line length to a nice round number

zachmullen:
This is a very minor thing, but the previous string construction included all the white space from the second line, which is presumably not what was intended. A better way to handle strings that don't fit on one line is to use lexically adjacent literals:

```python
(
    'This is the first line, '
    'and this is the second.'
)
```


mvandenburgh (dismissed):

rdwatch/core/models/region.py
```diff
@@ -83,8 +83,7 @@ class Meta:
                 name='unique_region_with_owner',
                 fields=['name', 'owner'],
                 condition=Q(owner__isnull=False),
-                violation_error_message='Region already exists with \
-                    this name and owner.',
+                violation_error_message='Region already exists with this name and owner.',
```
mvandenburgh, line 86: `flake8` is reporting this line as too long.

mvandenburgh, line 86: I missed that this is in a model. So we need a new migration for this

mvandenburgh (approved):

PR #490, opened 2024-08-15
Support uploading model runs from the web client

floryst:
Add support for uploading model runs from the web client.

This draft PR gets the end-to-end pipeline of uploading and processing a zip file of geojsons. There are a couple of extra things to do to complete this PR.
- [x] Cleanup the UI
- [x] Enforce certain validations on the UI (e.g. not allowed to submit without a title)
- [x] Send the private flag, and apply the private flag on the backend
- [x] Run code linter
- [x] Address the TODOs

The core pipeline can use an initial review pass.

Screenshots:
image

image

image


floryst:
@BryonLewis is there an existing notion of a private/public model run in RD-WATCH?

rdwatch/core/tasks/__init__.py
```diff
@@ -719,3 +724,99 @@ def download_sam_model_if_not_exists(**kwargs):
...
+        )
+        model_run = ModelRun.objects.create(
+            title=model_run_upload.title,
+            performer=performer,
+            region=region,
+            # TODO toggle for {'ground_truth': True/False}?
+            # TODO is this where the private toggle can be used?
+            parameters={},
+            # TODO handle expiration_time, evaluation, evaluation_run, proposal?
+            # TODO pass in the user?
```
floryst, line 805: @BryonLewis I'm defaulting to not adding these arguments, but I left them here to get your opinion on it.

vue/src/components/SideBar.vue
```diff
@@ -183,6 +184,14 @@ const satelliteLoadingColor = computed(() => {
           <span class="text-caption">{{ state.appVersion }}</span>
         </div>
       </v-row>
+      <v-row
+        dense
+        align="center"
+        justify="center"
+        class="mb-3"
+      >
+        <upload-model-run />
```
BryonLewis, line 193: on success should a signal be emitted that indicates the modelrun listing should be reloaded?

BryonLewis (changes requested):
I haven't tested yet, just looked at the code and added a few suggestions.

vue/src/components/UploadModelRun.vue
```diff
@@ -0,0 +1,196 @@
...
+async function upload() {
+  try {
+    uploadLoading.value = true;
+    uploadError.value = undefined;
+    uploadProgress.value = Infinity;
+
+    const file = Array.isArray(uploadFile.value)
+      ? uploadFile.value[0]
+      : uploadFile.value;
+    if (!file) return;
```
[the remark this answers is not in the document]
floryst: Vuetify is supposed to produce a `File` if `multiple` isn't set, and `File[]` if `multiple` is set (or the modelValue is an array) (https://github.com/vuetifyjs/vuetify/issues/17687). However, I'm consistently seeing an array show up for some unknown reason.

I think it's safer to treat the file ref as being either an array or a file object, given that always returning an array may in fact be buggy behavior. I will also push some changes to properly annotate the ref type, since I have it as `ref<File>()` currently.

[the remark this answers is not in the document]
BryonLewis: ahh thanks, I didn't search for an existing bug.

pyproject.toml
```diff
@@ -60,6 +60,7 @@ apache-airflow-client = "^2.9.0"
 beautifulsoup4 = "^4.12.3"
 django-allauth = {extras = ["socialaccount"], version = "^0.63.2"}
 django-login-required-middleware = "^0.9.0"
+django-s3-file-field = {extras = ["minio"], version = "^1.0.1"}
```
mvandenburgh, line 63: The `minio` variant should only be installed in development. In production, we'll need the `s3` variant.

You can run these commands to get poetry to do it:

```bash
poetry remove django-s3-file-field
poetry add django-s3-file-field --extras "s3"
poetry add django-s3-file-field --group dev --extras "minio"
```

vue/src/components/SideBar.vue
floryst: I added a commit that refreshes the performers and regions. Is there another bit of state that should be refreshed when the model run is uploaded?

floryst:
Recent updates:
* use django-s3-file-field with s3 in prod, minio in dev
* properly type the VFileInput model value
* skip `originator` validation in SiteFeature. This is necessary because ingested geojsons may have originators that are not yet in the database. We can safely skip this check because we will add the missing performers later before saving the model run.
  * side note: the implementation would be better with pydantic v2 contexts
* refresh the performers and regions when a successful upload occurs

vue/src/components/UploadModelRun.vue
```diff
@@ -0,0 +1,200 @@
...
+      }
+    }
+
+    queryTask();
+  });
+}
+
+async function upload() {
+  try {
+    uploadLoading.value = true;
```
BryonLewis, line 65: Can we use this to disable the inputs while the upload is processing?

There are some notes further down about the dialog being persistent and if we could allow multiple uploads processing at once in the celery tasks.  I think for now the additional code to do that is a bit much so just disabling the inputs when the upload is processing is enough.

BryonLewis (changes requested):
Sorry I've been a bit busy on other projects but got around to review this.  Some of the comments are generic things that could be implemented in the future.  The main things are:
- The Region override not working properly
- I think it would be good to get the modelrun list to refresh on upload task completion.

The section about allowing multiple uploads to be processed at once and having the dialog be non-blocking isn't as important now considering modelruns should be processed in a few seconds.

rdwatch/core/tasks/__init__.py
```diff
@@ -719,3 +724,107 @@ def download_sam_model_if_not_exists(**kwargs):
...
+            public=not model_run_upload.private,
+        )
+
+        for site_model in site_models:
+            SiteEvaluation.bulk_create_from_site_model(site_model, model_run)
+        for region_model in region_models:
+            SiteEvaluation.bulk_create_from_region_model(region_model, model_run)
```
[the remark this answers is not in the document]
floryst: I'm making some changes to handle this case.
- Uploaded zip files must have only 1 valid region model
- If there is an override region ID, then the region model's RegionFeature.properties.region_id is overwritten before it gets sent to `bulk_create_from_region_model`.

I think that should handle the case above without having to add additional parameters to that function.

vue/src/components/SideBar.vue
```diff
@@ -159,8 +160,10 @@ const satelliteLoadingColor = computed(() => {
   return 'black'
 })
 
-
-
+const refreshListings = () => {
+  updatePerformers();
+  updateRegionList();
```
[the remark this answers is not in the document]
floryst: How about merging `updateRegionMap` into `updateRegionList`? I see that these two functions are always called together, so I can move `updateRegionMap` into `updateRegionList`, keeping the name `updateRegionList`.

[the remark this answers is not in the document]
BryonLewis: Go for it, if they are always called together that should be fine.  I really don't remember why I separated them out.  I had something to do with custom Regions but it may have been just for that PR.

rdwatch/core/tasks/__init__.py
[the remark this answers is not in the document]
mvandenburgh: I like the approach @floryst added :+1:

rdwatch/core/views/model_run.py
```diff
@@ -637,3 +650,35 @@ def get_downloaded_annotations(request: HttpRequest, id: UUID4, task_id: str):
...
+            performer=upload_data.performer.strip(),
+            zipfile=zipfile_upload['object_key'],
+            private=upload_data.private,
+        )
+        if not upload.title:
+            raise ValidationError('Invalid model run title')
+        if not default_storage.exists(upload.zipfile.name):
+            raise ValidationError('Invalid file name provided')
+
+        task = process_model_run_upload_task.delay(upload.id)
```
mvandenburgh, line 674: Theoretically (although unlikely in practice), the transaction creating the `ModelRunUpload` could still be running when the celery task starts -

```suggestion
        task = process_model_run_upload_task.delay_on_commit(upload.id)
```

mvandenburgh (changes requested):

vue/src/components/UploadModelRun.vue
floryst: Done!

vue/src/components/SideBar.vue
```diff
@@ -159,8 +160,10 @@ const satelliteLoadingColor = computed(() => {
   return 'black'
 })
 
-
-
+const refreshListings = () => {
```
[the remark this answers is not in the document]
floryst: The second option is now in #503

rdwatch/core/views/model_run.py
floryst: So `delay_on_commit` doesn't return a task ID. I'm thinking of updati this code to instead save the upload task, call `delay_on_commit()`, and have the task itself write its ID to the upload instance. Instead of returning the task ID, we will return the upload instance ID, and have that query the task status.

rdwatch/core/tasks/__init__.py
```diff
@@ -762,3 +767,118 @@ def download_sam_model_if_not_exists(**kwargs):
...
+    model_run_upload = ModelRunUpload.objects.get(pk=upload_id)
+
+    try:
+        with transaction.atomic():
+            model_run_upload.task_id = task.request.id
+            model_run_upload.save()
+
+        process_model_run_upload(model_run_upload)
+    finally:
+        model_run_upload.delete()
```
mvandenburgh, line 884: I would pessimistically lock this if you are using .save(), since if the `model_run_upload` object becomes stale between the `.get` and `.save` call, the UPDATE query generated from .save() can clobber the row. 

```suggestion
    try:
        with transaction.atomic():
            model_run_upload = ModelRunUpload.objects.select_for_update().get(pk=upload_id)
            model_run_upload.task_id = task.request.id
            model_run_upload.save()

        process_model_run_upload(model_run_upload)
    finally:
        model_run_upload.delete()
```

Or better yet, just reduce it to a single UPDATE query and avoid the staleness question entirely
```suggestion
    ModelRunUpload.objects.filter(pk=upload_id).update(task_id=task.request.id)

    try:
        process_model_run_upload(ModelRunUpload.objects.get(pk=upload_id))
    finally:
        model_run_upload.delete()
```

mvandenburgh (changes requested):

rdwatch/core/tasks/__init__.py
floryst: Agreed! Thanks for catching that. The single update query is what I'm looking to do.

rdwatch/core/tasks/__init__.py
```diff
@@ -762,3 +767,114 @@ def download_sam_model_if_not_exists(**kwargs):
...
+
+
+@shared_task(bind=True)
+def process_model_run_upload_task(task, upload_id: UUID):
+    ModelRunUpload.objects.filter(pk=upload_id).update(task_id=task.request.id)
+
+    try:
+        process_model_run_upload(ModelRunUpload.objects.get(pk=upload_id))
+    finally:
+        model_run_upload.delete()
```
BryonLewis, line 880: After the last update this needs to be modified.

BryonLewis (approved):

mvandenburgh (approved):

BryonLewis (approved):
Tested on a full unique zip with custom regions and size and it worked perfectly.  Even the overrides worked well.

PR #492, opened 2024-08-23
Enable Ground Truth List for Simple/Detailed views

BryonLewis:
Previously the Site List while in scoring mode and looking at the simple or detailed views wouldn't display the ground truth in the list.

Updates:

- Makes sure the simple/detailed scoring views don't filter out the ground truth
- Adds to the backend the color_code data for each site.  This may be an expensive operation in comparison to not having it
- Handles filtering the sites in the list based on the simple/detailed display properties in the annotation colors
- Adds an X/Y (x out of y) to the site list to indicate how many sites are displayed out of the total number of sites.

https://github.com/user-attachments/assets/aab4a00c-b454-4e5d-9bf6-7dab45bae282

### DIFF
--- a/rdwatch/core/utils/raster_tile.py
+++ b/rdwatch/core/utils/raster_tile.py
@@ -9,6 +9,7 @@ logger = logging.getLogger(__name__)
 
 def get_raster_tile(uri: str, z: int, x: int, y: int) -> bytes:
     # logger.warning(f'SITE URI: {uri}')
+    logger.warning(f'Image URI: {uri}')
     with rasterio.Env(GDAL_DISABLE_READDIR_ON_OPEN='EMPTY_DIR'):
         if uri.startswith('https://sentinel-cogs.s3.us-west-2.amazonaws.com'):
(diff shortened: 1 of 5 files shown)


rdwatch/core/utils/raster_tile.py
```diff
@@ -9,6 +9,7 @@
 
 def get_raster_tile(uri: str, z: int, x: int, y: int) -> bytes:
     # logger.warning(f'SITE URI: {uri}')
+    logger.warning(f'Image URI: {uri}')
```
mvandenburgh, line 12:
```suggestion
    # logger.warning(f'SITE URI: {uri}')
```

Or if we want to keep this log, I think we should remove the commented line.

BryonLewis: No reason to keep it, it was when I was debugging the issues with the downloading and some images not coming through with the removal of stuff form the S3.

mvandenburgh (approved):
Thanks for the video explanation, that helped clarify this. I imagine the performance bottleneck is still the vector tile generation itself, so I think having the new subquery for color code isn't too bad.

PR #493, opened 2024-08-28
Element84 STAC Query Support

BryonLewis:
This adds support for Element84 STAC Query server for Sentinel 2 imagery.  https://earth-search.aws.element84.com/v1/

NOTES:

- This updated version only works for S2 Imagery.  There is a collection for Landsat8 but in my quick testing it didn't find any images.
- The new images require bit scaling instead of the SMART programs 0-10000 scaling, so the default scaling mechanism is now 'bits' to calculate the value
- There are no longer images directly tied to observations, so test data that relies on observations may not have overlapping imagery.  A quick test shows that it has image data for more recent timeframes.  I.E searching for images between 2016-2019 will produce less images than searching from 2022-2024.

INITIALIZATION:
This requires copying over the environment variable: `RDWATCH_STAC_URL="https://earth-search.aws.element84.com/v1/"` from the `template.env` to your local `.env` to work.

Technical Changes:

- `RDWATCH_STAC_URL` environment variable was added to the system.  This is for S2/L8 imagery right now.  When this environment variable is present for the S2/L8/PL it will use this STAC server instead of the SMART_STAC_URL server. 
- Using settings I add collections for the SMART_STAC_URL server or for the STAC_URL server.  So collections change based on the presence of the RDWATCH_STAC_URL environment variable.
- I support returning empty values if a collection doesn't exists for a certain source.  I.E when using the new RDWATCH_STAC_URL` and `PL` it will return an empty list because there is no collection added for that source.
- The satellite_bands.py would process the results of the STAC query and indicate the processing level based on the collection that was used.  In modern times (last 1.5 years) we have only used the 2A processing level because all of our collections have been 2A.  I've updated the case conditional to reflect that it needs to find one of the values in the current Collection and will set the level to 2A.
- Defaulted all scales for image scaling to 'bits' meaning it calculates the low and high levels of the image and scales accordingly.  We can't rely on the SMART 0-10,000 scaling anymore.
- Previously the worldviewSource was the only thing guarding a bunch of worldview only functions.  I've updated it to make sure it checks for `WV` as the constellation type to be a bit more secure.  This also allows in the future to rename `worldviewSource` if we want to support COG/NITF for other constellations
-


rdwatch/core/utils/stac_search.py
```diff
@@ -56,10 +56,18 @@ def _fmt_time(time: datetime):
     'PL': [],
 }
 
-if settings.ACCENTURE_VERSION is not None:
+STAC_URL = ''
+STAC_HEADERS = {}
+if settings.STAC_URL:
+    COLLECTIONS['S2'].append(['sentinel-2-c1-l2a', 'sentinel-2-l2a'])
```
mvandenburgh, line 62:
```suggestion
    COLLECTIONS['S2'] += ['sentinel-2-c1-l2a', 'sentinel-2-l2a']
```

mvandenburgh (changes requested):

mvandenburgh (approved):
I tested this locally and it works. @BryonLewis PTAL at the backend changes I pushed and merge if they look ok

mvandenburgh (approved):
Works for me :+1:

PR #501, opened 2024-09-11
Combine raster image bands if multiple are available

mvandenburgh:
This updates the logic for pulling S2 raster imagery so that if the image contains multiple bands, all of them are combined prior to rendering.

See https://cogeotiff.github.io/rio-tiler/readers/#rio_tileriorasterioreader for example.


floryst:
I've moved this over to #510 (https://github.com/ResonantGeoData/RD-WATCH/pull/510/files#diff-47d30659c222eac3045105d112673bb128ab41c2238b04db41a7f1f38ba5bc1eR40).

PR #502, opened 2024-09-11
Fix ModelRunList styles and pagination

floryst:
- The pagination wasn't working properly, as `limit` was being sent rather than `page` and the server is using PageNumberPagination. This meant that triggering pagination would result in the same model run list being sent for each request.
- The styles for the model run list caused it to slightly overflow the bottom of the sidebar, resulting in some cut-off. This becomes more evident if elements are added to the top of the sidebar. Flexbox + overflow handling fixes this issue.

These are issues I discovered as I was moving the `loadModelRuns` function out of the ModelRunList component and into the store. I'll be looking to add cleaner pagination handling when I do so.

### DIFF
--- a/rdwatch/core/views/model_run.py
+++ b/rdwatch/core/views/model_run.py
@@ -54,6 +54,8 @@ from rdwatch.core.tasks import (
 from rdwatch.core.views.performer import PerformerSchema
 from rdwatch.core.views.site_observation import GenerateImagesSchema
 
+MODEL_RUN_PAGE_SIZE = 10
+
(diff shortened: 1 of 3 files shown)


floryst:
The failing test (test_list_model_run_rest_list) is failing because we are now returning a page size of 10 instead of the default 100. Do we want `GET /model-runs/` to return a page size of 10 or 100 (or another value)? I kept with 10 because that's what the ModelRunList.vue component wanted with `const limit = 10`.

mvandenburgh (approved):
LGTM

PR #503, opened 2024-09-12
Refactor loadModelRuns to the store as a global action

floryst:
- `loadModelRuns` is now part of the store as `queryModelRuns` and updates store state.
- `queryModelRuns` manages pagination state through "firstPage" and "nextPage" requests (see ModelRunList.vue for usage).
- `checkDownloading` is now part of the store as `refreshModelRunDownloadingStatuses`. Additionally, it has been fixed to properly update the downloading statuses of all loaded pagination pages.
  - A side note: I have yet to test this function!
- All usages of `state.downloadCheck` have been replaced with calls to `refreshModelRunDownloadingStatuses`.

### DIFF
--- a/vue/src/components/siteList/SiteList.vue
+++ b/vue/src/components/siteList/SiteList.vue
@@ -10,7 +10,7 @@ import SiteListCard from "./SiteListCard.vue";
 import { SiteDisplay } from "./SiteListCard.vue";
 import SiteListHeader from "./SiteListHeader.vue";
 import SiteListFilter from "./SiteListFilter.vue";
-import { getSiteObservationDetails, state } from "../../store";
+import { getSiteObservationDetails, refreshModelRunDownloadingStatuses, state } from "../../store";
 import { scoringColors, scoringColorsKeys } from "../../mapstyle/annotationStyles";
 
 const props = defineProps<{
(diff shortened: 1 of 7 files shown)


floryst:
This technically conflicts with #502, but it's still reviewable in its current form. This should be merged into/after #502.

mvandenburgh (commented):
This makes sense to me, and it works without any issues. I'll defer to @BryonLewis for approval since this is a clientside-only change

vue/src/components/siteList/SiteList.vue
```diff
@@ -283,12 +283,12 @@ const startDownload = async (data: DownloadSettings) => {
...
-  if (imageDownloadingModelRunId.value) {
-    emit('image-download', imageDownloadingModelRunId.value);
-  }
-  state.downloadingCheck += 1;
+    await ApiService.getObservationImages(id, data);
+    // Notify the ModelRunList that downloading is happening
+    if (imageDownloadingModelRunId.value) {
+      emit('image-download', imageDownloadingModelRunId.value);
+    }
+    refreshModelRunDownloadingStatuses();
```
mvandenburgh, line 291: I assume this was autoformatted. I'll file an issue to investigate why the linting step didn't flag this the first time around

floryst: I think the linting step isn't performing as intended when I run `npm run lint:fix`, but I haven't looked into it yet.

BryonLewis (approved):
Thanks Again this was extremely helpful.

PR #504, opened 2024-09-12
Cleaner version display format

mvandenburgh:
Updates the version string in the UI to be more user friendly.

Tagged release versions now look like this:
Screenshot from 2024-09-12 11-55-10

And pre-release versions remain the same:
Screenshot from 2024-09-12 11-56-39

### DIFF
--- a/rdwatch/core/views/server_status.py
+++ b/rdwatch/core/views/server_status.py
@@ -3,6 +3,7 @@ import socket
 from typing import Any
 
 from ninja import Field, Router, Schema
+from parver import Version
 
 from django.http import HttpRequest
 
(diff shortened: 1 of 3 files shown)


floryst (approved):
LGTM!

As a very initial thought, if we want to avoid adding another dependency, it might be sufficient to use a regex to match against the release part of the version string: `^v[0-9]+(\.[0-9]+)*`

rdwatch/core/views/server_status.py
```diff
@@ -40,6 +41,16 @@ def get_status(request: HttpRequest):
...
 
+    parsed_version = Version.parse(api.version)
+
+    version = 'v'
+    version += '.'.join(str(v) for v in parsed_version.release)
+
+    # If this is a pre-release version, include the pre-release version string
+    post_num: int | None = parsed_version.post
+    if post_num is not None:
+        version += f'.post{post_num}+{parsed_version.local}'
```
floryst, line 52: What's the difference between `api.version` and the string constructed here?

mvandenburgh: When the current version is not a tagged release, they differ only slightly. For example, for the latest commit on this branch
```python
print(api.version)  # prints '1.10.2.post5.dev0+b074e39c'
print(version)  # prints 'v1.10.2.post5+b074e39c'
```

The main motivation behind this change is to make it look more user friendly when the curent version is a tagged release, which looks like this
```python
print(api.version)  # prints '1.12+b074e39c'
print(version)  # prints 'v1.12'
```

PR #506, opened 2024-09-13
set timeMin to modelRun list timeMin

BryonLewis:
The removal of the `@update:timeRang` from `SideBar.vue` meant that the slider was always going to utilize the UNIX epoch time.  This uses the global `state.timeMin`.  I think this was an oversight of mine when I was converting the very heavily integrated component system signaling (where it was doing some passing up/down of information) into more of a store system when I came on to the project.

I do understand that our Store is really overloaded right now and at some point should probably be refactored but that is an issue for another time.

### DIFF
--- a/vue/src/components/SideBar.vue
+++ b/vue/src/components/SideBar.vue
@@ -16,7 +16,6 @@ import { changeTime } from "../interactions/timeStepper";
 import { useRoute } from "vue-router";
 import ModeSelector from './ModeSelector.vue';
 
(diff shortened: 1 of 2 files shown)


floryst:
Does the same change need to happen to `annotationViewer/SideBar.vue`? There's also a timemin change there.

BryonLewis:
> Does the same change need to happen to `annotationViewer/SideBar.vue`? There's also a timemin change there.

It was sitting in my commit list  in VSCode (I guess I forgot to click on commit all).  Thanks,

PR #507, opened 2024-09-13
fix modelrun animation creating a file for each image

BryonLewis:
Noticed an issue with the orevious animation tasks where it was generating an AnimationSiteExport for each individual image in a site instead of checking to see if a site has 1 or more images.  This resulted in it sometimes creating hundreds of files based on the number of sites that were available.


rdwatch/core/tasks/animation_export.py
```diff
@@ -746,10 +746,12 @@ def create_modelrun_animation_export(
     site_tasks = [
         create_site_animation_export.s(site_id, settings, userId)
         for site_id in SiteImage.objects.filter(site__configuration_id=modelrun_id)
-        .values_list('site_id', flat=True)
-        .iterator()
+        .values('site_id')  # Group by site_id
+        .annotate(image_count=Count('id'))  # Count images per site_id
```
mvandenburgh, line 750: I haven't tested yet, but shouldn't this be

```suggestion
        .annotate(image_count=Count('site_id'))  # Count images per site_id
```

?

mvandenburgh (approved):

mvandenburgh (approved):
LGTM, just need linting fix

mvandenburgh (approved):

PR #508, opened 2024-09-18
Scoring Task Parity

BryonLewis:
Done:
- Update Image Downloading tasks to support both points (requires having point_geometry in Site Model) as well as the worldview nitf file support.
- Remember point annotations are only groundtruth sites, observations don't seem to have points
- Disabled the "upload model run" button in the scoring mode because it isn't supported.
- Created the AnimationSiteExport and AnimationModelRunExport models for the scoring database and applied migrations
    - This required updating the database router to allow some relations to the 'User' field
- Added animation_export.py task and updated it to reference the Scoring Database.
    - Need to fall back between Proposals and main sites because unlike the RGD database they aren't the same
    - Scoring database images are stored in 3857 and don't need transformation of the coordinates like RGD
- Added 'Cancel' endpoint for AnimationDownloads to both RGD and Scoring endpoints
- Added Cancel logic to the client
- properly clearing the timeout for checking download status when unMounting the download status page.
- Updated client for refreshing siteList so that it always properly filters the groundtruth.  Before when selecting it would display groundTruth even when it wasn't selected.
- Added some documentation about ingesting point scoring
- @floryst the model-run updating after downloading images was attempting to load page -1 in the scoring DB, I've updated the code to fix this.  Please take a look at it and make sure it makes sense, I didn't fully look into the nextPage feature.

KNOWN ISSUE:

- Point rescaling downloads is not placing the point in the proper location.  This is a known thing that can be addressed in the future


rdwatch/core/tasks/__init__.py
```diff
@@ -318,6 +318,11 @@ def get_siteobservations_images(
         timestamp = (min_time - timedelta(days=30)) + timebuffer
 
     # Now we get a list of all the timestamps and captures that fall in this range.
+    logger.warning('MAXBBOX')
+    logger.warning(max_bbox)
+    logger.warning('timestamp')
+    logger.warning(timestamp)
+    logger.warning(timebuffer)
```
floryst, line 325: Are these debugging lines?

vue/src/store.ts
```diff
@@ -635,13 +635,11 @@ async function queryModelRuns(type: 'firstPage' | 'nextPage', filters: QueryArgu
...
 const refreshModelRunDownloadingStatuses = async () => {
   const queryState = state.queryStates.loadModelRuns;
-  const loadedPages = queryState.nextPage - 1;
   const activeFilters = queryState.activeFilters;
-
   let existingRequest = queryState.request as CancelablePromise<void> | undefined;
 
   // chain cancelable requests
-  for (let page = 0; page < loadedPages; page++) {
+  for (let page = 1; page < queryState.nextPage; page++) {
```
floryst, line 642: Yep, I have this fixed in #509, but I like using `queryState.nextPage` instead (I was initially concerned with race conditions updating it while this task ran, but I'm less worried about that now.) I'll remove it from my PR and have it updated here.

rdwatch/core/tasks/__init__.py
BryonLewis: Yeah, I'll remove them.  I might leave some on that help with debugging issues l.

rdwatch/core/migrations/0036_modelrunupload_alter_animationmodelrunexport_user_and_more.py
```diff
@@ -56,4 +59,22 @@ class Migration(migrations.Migration):
...
+        ),
+        migrations.AlterField(
+            model_name='animationsiteexport',
+            name='user',
+            field=models.ForeignKey(
+                on_delete=django.db.models.deletion.CASCADE,
+                related_name='core_animationsiteexports',
+                to=settings.AUTH_USER_MODEL,
+            ),
+        ),
```
mvandenburgh, line 79: Shouldn't this be in a new migration, instead of modifying an existing one? Otherwise, this won't apply properly to deployments that have already has migration 36 applied.

rdwatch/core/views/animation.py
```diff
@@ -181,3 +182,22 @@ def get_animation_status(request: HttpRequest, task_id: UUID4):
...
+def cancel_animation_status(request: HttpRequest, task_id: UUID4):
+    try:
+        export_record = AnimationSiteExport.objects.get(celery_id=task_id)
+    except AnimationSiteExport.DoesNotExist:
+        export_record = AnimationModelRunExport.objects.get(celery_id=task_id)
+    if export_record:
+        with transaction.atomic():
+            # Use select_for_update here to lock the AnimationExport row
+            # for the duration of this transaction in order to ensure its
+            # status doesn't change out from under us
```
mvandenburgh, line 197:
```suggestion
```

rdwatch/scoring/models/taskexport.py
```diff
@@ -0,0 +1,18 @@
...
+
+    arguments = models.JSONField(null=True, default=None)
+    user = models.ForeignKey(User, on_delete=models.CASCADE)
+
+
+class AnimationModelRunExport(AbstractFileExport):
+    configuration = models.CharField(max_length=255)
+
+    user = models.ForeignKey(User, on_delete=models.CASCADE)
+    arguments = models.JSONField(null=True, default=None)
```
mvandenburgh, line 18: To maintain the mirrored file hierarchy between the `core` and `scoring` apps, could this module be named `annotation_exports.py`?

rdwatch/scoring/views/animation.py
```diff
@@ -0,0 +1,210 @@
...
+    logger.warning(f'Found: {exports.exists()} with id: {id}')
+    return generate_download_data(exports)
+
+
+class DeleteErrorSchema(BaseModel):
+    error: str
+
+
+class DeleteSuccessSchema(BaseModel):
+    success: str
```
mvandenburgh, line 105: Instead of redefining these, can we also import them from the `core` app like the other schemas? They're basic classes, but I think it's still worth being consistent and it also enforces consistency between the two APIs.

```python
from rdwatch.core.views.animation import DeleteErrorSchema, DeleteSuccessSchema
```

rdwatch/scoring/views/site_observation.py
```diff
@@ -35,13 +35,21 @@
...
     scale: Literal['default', 'bits', 'custom'] = 'bits'
     scaleNum: None | list[int] = None
     bboxScale: None | float = 1.2
+    pointArea: None | float = 200
+
+    @root_validator
+    def validate_worldview_source(cls, values: dict[str, Any]):
+        if 'WV' in values['constellation'] and values['worldviewSource'] is None:
+            raise ValueError('worldviewSource is required for WV constellation')
+        return values
```
mvandenburgh, line 52: Didn't we remove this because it was causing problems? I don't recall what the exact issue was, though.

mvandenburgh (changes requested):

docs/DevDocs/IngestingData.md
```diff
@@ -62,6 +62,26 @@ To score data:
...
+               --no-viz \
+               --no-viz-detection-table \
+               --no-viz-comparison-table \
+               --no-viz-associate-metrics \
+               --no-viz-activity-metrics \
+               --sequestered_id KR_R002 \
+               --db_conn_str postgresql+psycopg2://scoring:secretkey@localhost:5433/scoring \
+               --rm_dir ../KR_R002/region_models/ \
+               --sm_dir ../KR_R002/site_models/ \
+               --gt_points_file ../annotations/supplemental_data/point_based_annotations.geojson
```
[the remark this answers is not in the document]
BryonLewis: It's in master now, so I think a pull of the repo should have it in the referenced location.  The other example references the ./annotations/site-models directory.  If you have that and an up to date annotations repo you should have that file.

rdwatch/core/migrations/0036_modelrunupload_alter_animationmodelrunexport_user_and_more.py
BryonLewis: Good catch, I think it got it merged into the updates for the modelrunupload.  I'll make it it's own migration.

rdwatch/scoring/models/taskexport.py
BryonLewis: I think I'll rename the core one task_exports.py.  It's really more than just annotation_exports now.  It has the base class for exporting annotations and animations.

rdwatch/scoring/views/site_observation.py
BryonLewis: https://github.com/ResonantGeoData/RD-WATCH/pull/472#discussion_r1709952201

So there was a default version located above.
Then I modified it in the STAC Query PR. 
https://github.com/ResonantGeoData/RD-WATCH/pull/493/files#diff-77c78040069f3371d705c88e756aefd249aa57d2b0d893c699d8d6ab58faad63

The update allows you provide a worldview-source (NITF/COG) even if you don't have WV selected.  So that way we can have a global default value.  I'm using the same one as the core version.

rdwatch/scoring/tasks/animation_export.py
```diff
@@ -0,0 +1,698 @@
...
+            )  # TODO Check if it needs to change _isempty or something like that
+        if 'nonobs' in filters['include'] and len(filters['include']) == 1:
+            query &= Q(
+                observation__isnull=True
+            )  # TODO check if it needs to change _isempty or something like taht
+
+    images = SiteImage.objects.filter(query).order_by('timestamp')
+
+    if len(images) == 0:
+        logger.warning('No Images found returning')
```
mvandenburgh, line 197:
```suggestion
        logger.debug('No Images found returning')
```

rdwatch/scoring/tasks/animation_export.py
```diff
@@ -0,0 +1,698 @@
...
+        return False, False
+    total_images = len(images)
+    max_image_record = None
+    max_width_px, max_height_px = 0, 0
+    for image_record in images:
```
mvandenburgh, line 202:
```suggestion
    for image_record in images.iterator():
```

rdwatch/scoring/tasks/animation_export.py
```diff
@@ -0,0 +1,698 @@
...
+        .first()
+    )
+
+    return closest_observation
+
+
+@shared_task
+def create_animation(
+    self, site_evaluation_id: UUID4, settings: GenerateAnimationSchema
+):
```
mvandenburgh, line 141: This is something that also affects the existing `core` create_animation task, but I didn't notice it until now - does this actually need to be a `shared_task`? It's only ever called from another Celery task, and more importantly, doesn't work as a Celery task due to it returning a file path local to the worker it was run on. 

```suggestion
def create_animation(
    self, site_evaluation_id: UUID4, settings: GenerateAnimationSchema
):
```

rdwatch/scoring/views/animation.py
```diff
@@ -0,0 +1,203 @@
...
+def cancel_animation_status(request: HttpRequest, task_id: UUID4):
+    try:
+        export_record = AnimationSiteExport.objects.get(celery_id=task_id)
+    except AnimationSiteExport.DoesNotExist:
+        export_record = AnimationModelRunExport.objects.get(celery_id=task_id)
+    if export_record:
+        with transaction.atomic():
+            # Use select_for_update here to lock the AnimationExport row
+            # for the duration of this transaction in order to ensure its
+            # status doesn't change out from under us
```
mvandenburgh, line 197:
```suggestion
```

rdwatch/scoring/tasks/animation_export.py
```diff
@@ -0,0 +1,698 @@
...
+        .order_by('-date')
+        .first()
+    )
+
+    return closest_observation
+
+
+@shared_task
+def create_animation(
+    self, site_evaluation_id: UUID4, settings: GenerateAnimationSchema
```
mvandenburgh, line 140: The `settings` object is not an instance of `GenerateAnimationSchema`, but a `GenerateAnimationSchema` that is serialized to a `dict`.

```suggestion
    self, site_evaluation_id: UUID4, settings: dict[str, Any]
```

rdwatch/scoring/tasks/animation_export.py
```diff
@@ -0,0 +1,698 @@
...
+            },
+        )
+
+        return output_file_path, name
+    return False, False
+
+
+@app.task(bind=True)
+def create_site_animation_export(
+    self, site_evaluation_id: UUID4, settings: GenerateAnimationSchema, userId: int
```
mvandenburgh, line 582: Similar to above issue - 
```suggestion
    self, site_evaluation_id: UUID4, settings: dict[str, Any], userId: int
```

rdwatch/scoring/tasks/animation_export.py
```diff
@@ -0,0 +1,698 @@
...
+        os.remove(file_path)
+    # except Exception as e:
+    #     logger.warning(f'Error when processing Animation: {e}')
+    #     if site_export:
+    #         site_export.delete()
+
+
+@app.task(bind=True)
+def create_modelrun_animation_export(
+    self, modelrun_id: UUID4, settings: GenerateAnimationSchema, userId: int
```
mvandenburgh, line 620: Similar to above issue -
```suggestion
    self, modelrun_id: UUID4, settings: dict[str,  Any], userId: int
```

mvandenburgh (changes requested):
I noticed a couple other minor issues. The one that is also present in the `core` animation task, I think we could also just postpone to another PR if needed.

rdwatch/scoring/tasks/animation_export.py
BryonLewis: Yup it probably started as a shared task until I redesigned it to abstract out some things with the path.

rdwatch/scoring/tasks/animation_export.py
BryonLewis: Yup celery was passing it as that.

mvandenburgh (approved):
LGTM. I tested locally and it works as expected :+1:

PR #509, opened 2024-09-18
Improve upload flow

floryst:
- Report errors to the client
- allow for zip files to not have region models if an override region is specified
- list regions and performers in their respective dropdowns in the upload box

### DIFF
--- a/rdwatch/core/tasks/__init__.py
+++ b/rdwatch/core/tasks/__init__.py
@@ -805,50 +805,66 @@ def process_model_run_upload(model_run_upload: ModelRunUpload):
 
     if len(site_models) == 0:
         raise ValueError('Did not receive any site models')
-    if len(region_models) == 0:
-        raise ValueError('Did not receive any region models')
-    if len(region_models) > 1:
-        raise ValueError('Too many regions provided in the zip file')
 
-    region_model = region_models[0]
-    region_feature = region_model.region_feature
-    region_id = region_feature.properties.region_id
-    region_originator = region_feature.properties.originator
+    override_region = model_run_upload.region
+    override_performer = model_run_upload.performer
 
-    # validate site models against the single region model
+    if override_region:
+        region_id = override_region
+        region_model = None
+        region_geometry = None
+        region_originator = None
+
+        # apply overrides
+        for model in site_models:
+            model.site_feature.properties.region_id = override_region
+    else:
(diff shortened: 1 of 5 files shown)


floryst:
New changes:
* fixed pagination off-by-one
* full upload error tracebacks are sent to the app-level error handler
* upload button hidden in annotator view

mvandenburgh (approved):
LGTM :+1:

BryonLewis (changes requested):
Push back if you believe this is out of scope for this PR:

Proposal/Annotator Mode Upload Model Run:  Either removing the upload button or supporting the upload button with a default that the uploaded model is a 'proposal'.

Background:

The annotator mode used to be called 'proposal' so you'll see that referenced a lot in the code.
https://github.com/ResonantGeoData/RD-WATCH/blob/main/vue/src/router.ts#L16
The above router will take you into proposal mode.
https://github.com/ResonantGeoData/RD-WATCH/blob/main/vue/src/components/LayerSelection.vue#L11C1-L12C1
That code can determine if you are in the proposal/annotator mode by looking at the route to determine it.

Option 1 - Disabling Upload Model:
You could make the visibility of upload model run only occur when not in proposal mode and not in scoring mode here: https://github.com/ResonantGeoData/RD-WATCH/blob/main/vue/src/components/ModelRunList.vue#L192

Option 2 - supporting proposal uploads:
If you want to support proposal uploading, which may be handy in the future.  You would keep this in there for proposal mode and then edit the logic in the upload-model-run component to add an additional 'proposal' flag when loading the data.

As seen in the upload script:
https://github.com/ResonantGeoData/RD-WATCH/blob/main/scripts/loadModelRun.py#L120
The proposal is a boolean option when posting to the endpoint.

https://github.com/ResonantGeoData/RD-WATCH/blob/main/rdwatch/core/models/model_run.py#L46
The ModelRun model has a proposal option during creation where this can be added during your processing of the upload: https://github.com/ResonantGeoData/RD-WATCH/blob/main/rdwatch/core/tasks/__init__.py#L859
So just masking it so that upload-model-run vue component understands that it is currently being used in the proposal route and the upload should be processed as a proposal could work, or just disabling the button for now from the proposal route.

floryst:
Updated to hide the upload button in the annotator view.

BryonLewis (approved):

PR #510, opened 2024-09-25
L8 satellite images

floryst:
- `stac_search.py` and `satellite_bands.py`: Switch to using the pystac_client.ItemSearch interface directly
- `satellite_bands.py`: Modified `get_bands()` and the `Bands` type to be a STACCapture
- `raster_tile.py`: refactored get_raster_tile/bbox to accept a reader
- `raster_tile.py`: Removed "b1" key in favor of picking the first band
  - This is a point of discussion for multichannel images!
- `capture.py`: abstracted captures to have an `open_reader()` method for generic rio_tiler reader usage

### DIFF
--- a/rdwatch/core/utils/raster_tile.py
+++ b/rdwatch/core/utils/raster_tile.py
@@ -1,77 +1,120 @@
 import logging
+from contextlib import ExitStack
 from typing import Literal
 
 import rasterio  # type: ignore
+from pystac import Asset
 from rio_tiler.io.rasterio import Reader
+from rio_tiler.io.stac import STACReader
 
 logger = logging.getLogger(__name__)
+
+DEFAULT_RESCALE_RANGE = (1, 10000)
+
+
+def get_asset_num_bands(asset: Asset) -> int:
+    return len(asset.extra_fields.get('eo:bands', []))
+
+
+def get_read_kwargs_for_reader(reader: Reader | STACReader):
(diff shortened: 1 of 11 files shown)


floryst:
Actually, I think the colors are messed up. I need to make sure STACReader is setting the red/green/blue bands in the correct order.

rdwatch/core/utils/raster_tile.py
```diff
@@ -1,77 +1,120 @@
...
+    elif scale == 'bits':
+        low, high = get_rescale_range_from_reader(reader)
+        img.rescale(in_range=((low, high),))
+    return img.render(img_format='WEBP')
+
 
 def get_raster_tile(uri: str, z: int, x: int, y: int) -> bytes:
     # logger.warning(f'SITE URI: {uri}')
-    with rasterio.Env(GDAL_DISABLE_READDIR_ON_OPEN='EMPTY_DIR'):
+    with ExitStack() as cxt_stack:
```
mvandenburgh, line 75: I've never seen `contextlib.ExitStack` before, that's really nice!

floryst:
I think the colors are fixed now. More testing is needed to make sure.

BryonLewis (approved):
Pulled, tested it with some L8 images and it seemed to work properly.  Even exporting the images to Animations.

Thank you for abstracting that out and cleaning up some of the reading code.

PR #511, opened 2024-09-26
Change WARN logs to INFO where needed

mvandenburgh:
This is a prerequisite to configuring Sentry, since `WARNING` logs will set off an issue alert there. 

We may want to go back through and figure out if some of these can be removed, or at least downgraded to `DEBUG` logs. But this change will at least allow us to hook up our deployment to Sentry without blowing it up with issues.


BryonLewis (approved):
I'll have to update my Scoring Parity PR after this.  I may remove some of the MAX_BBOX stuff from the downloading tasks but will probably keep some number of images so that it makes some stuff easier while debugging locally.

PR #514, opened 2024-10-01
[wip] Fix satellite fetching race condition

### LINKED ISSUE #513: `SatelliteFetching` is created too late in image download process
Currently when a user kicks off an image download (for either a single site, or an entire model run), the endpoint simply places a message on the queue for the celery worker to pick up. Then, the [celery worker creates a `SatelliteFetching` row once the task starts](https://github.com/ResonantGeoData/RD-WATCH/blob/main/rdwatch/core/tasks/__init__.py#L616-L634) to signal that the image generation is running. This results in a race condition where, if the celery worker is preoccupied with another task or not up, the `SatelliteFetching` row will not be created. This means the UI will not update to let the user know the imagery task has started.

I can think of two ways to fix this -
1) Add a `PENDING` status to the `SatelliteFetching` model and set it in the endpoint prior to dispatching the task. Once the task begins, set it to `RUNNING`.
2) Just create the `SatelliteFetching` row in the endpoint with a status of `RUNNING` prior to dispatching the task.

I think option 2 makes more sense. Not only is it simpler to implement, but I don't see any value in distinguishing between a PENDING and RUNNING task in the UI. Either way the user is waiting, so it doesn't seem like it would offer any useful information.

mvandenburgh:
Fixes #513. 

This is only for the `core` app; once we settle on the correct way to fix this, I'll apply it to `scoring` as well.

### DIFF
--- a/rdwatch/core/tasks/__init__.py
+++ b/rdwatch/core/tasks/__init__.py
@@ -598,6 +598,7 @@ def cancel_generate_images_task(model_run_id: UUID4) -> None:
 @shared_task
 def generate_site_images(
     site_id: UUID4,
+    satellite_fetching_id: int,
     constellation=['WV'],  # noqa
     force=False,  # forced downloading found_timestamps again
     dayRange=14,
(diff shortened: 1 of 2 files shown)


BryonLewis:
I see that you updated the task or the downloading of modelRun images but the same problem then exists, where the modelrun download requires that the celery task be up and running before it is initiated and generates the SatelliteFetching rows.

This is going to require a change to: https://github.com/ResonantGeoData/RD-WATCH/blob/main/rdwatch/core/views/model_run.py#L511

mvandenburgh:
This might still be an issue, but this PR is very outdated at this point and will likely need to be redone. I won't be working on this for the foreseeable future, so I'll close this and we can open a new one when/if needed

PR #516, opened 2024-10-07
Construct true-color from R/G/B bands for S2 and L8

floryst:
- S2 now composites true-color from R/G/B bands rather than using the preprocessed true-color asset
- rescale each band individually to 0-255


BryonLewis:
would it make sense to change the default scaling in for S2 bands then?  Set it back to default instead of the bits (2%-98%)?

floryst:
I'm not yet sure of that. I think we need some more discussion around this before we even decide on this being a good route for the S2 images.

PR #517, opened 2024-10-14
Satellite Fetching Progress

BryonLewis:
- Updated Animation Generation for RDWATCH database to upscale when rescale is on for images that are lower than 500x500 pixels.  I changed it to use a 2X multiplier until both the width and height are above 500.  This makes it easier to draw the fonts and other items that are needed.
- Updated the satellite-fetching/running endpoint to return back the celery information that is logged during the image downloading progress.
- This image is now a hover card that shows up when hovering over the downloading status indicator.

TODO:
- [ ] Update animation creation for scoring to utilize the same upscaling option for lower resolution images
- [ ] Decide if I want to show more information for the downloading data.
    - [ ] Like showing what source it is on and has completed instead of just showing the current source.
    - [ ] Should I show the options for downloading?
- [ ] Test with a full model-run and make sure the performance isn't too bad with getting all the celery task information.  It may be that some sites are pending and I should indicate that while awaiting for them to start running

### DIFF
--- a/rdwatch/core/tasks/animation_export.py
+++ b/rdwatch/core/tasks/animation_export.py
@@ -372,9 +372,9 @@ def create_animation(self, site_evaluation_id: UUID4, settings: dict[str, Any]):
                 max_image_record = image_record
 
     if max_width_px < 500 or max_height_px < 500:
-        new_height = (max_height_px / max_width_px) * 500
-        max_width_px = int(500)
-        max_height_px = int(new_height)
+        while max_width_px < 500 or max_height_px < 500:
+            max_width_px *= 2
+            max_height_px *= 2
 
     # using the max_image_bbox we apply the rescale_border
     if rescale:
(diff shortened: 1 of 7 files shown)


rdwatch/core/tasks/animation_export.py
```diff
@@ -372,9 +374,9 @@ def create_animation(self, site_evaluation_id: UUID4, settings: dict[str, Any]):
                 max_image_record = image_record
 
     if max_width_px < 500 or max_height_px < 500:
-        new_height = (max_height_px / max_width_px) * 500
-        max_width_px = int(500)
-        max_height_px = int(new_height)
+        while max_width_px < 500 or max_height_px < 500:
```
floryst, line 377: The enclosing `if` condition can be removed.

rdwatch/scoring/tasks/animation_export.py
```diff
@@ -211,9 +211,9 @@ def create_animation(self, site_evaluation_id: UUID4, settings: dict[str, Any]):
                 max_image_record = image_record
 
     if max_width_px < 500 or max_height_px < 500:
-        new_height = (max_height_px / max_width_px) * 500
-        max_width_px = int(500)
-        max_height_px = int(new_height)
+        while max_width_px < 500 or max_height_px < 500:
```
floryst, line 214: The enclosing `if` condition can be removed

vue/src/components/siteList/SiteListCard.vue
```diff
@@ -359,13 +358,55 @@ const animationDialog = ref(false);
...
+            <v-card
+              v-if="downloadingData"
+              width="400"
+            >
+              <v-card-title><span style="font-size: 12px; font-weight:bold">{{ downloadingData.mode }}</span></v-card-title>
+              <v-card-text>
+                <v-row dense>
+                  <v-progress-linear
+                    :model-value="downloadingData.current"
+                    :max="downloadingData?.total"
```
floryst, line 376: Don't need optional chaining here with the `v-if` above.

floryst (commented):
Are there subtle-enough changes between core and scoring code-bases that de-duplication would be more effort than it's worth at this time?

floryst:
Haven't tested it, but overall LGTM

floryst (approved):

PR #518, opened 2024-10-14
Use POST when querying the STAC server

floryst:
The L8 STAC server returns incorrect pagination links when using GET search queries.


floryst:
As a follow-up: I have submitted a bug report to USGS. If they resolve the issue on their end, then I will close this PR.

BryonLewis (approved):

PR #519, opened 2024-10-15
Simplify scoring db checking

floryst:
Fix the ImageViewer incorrectly routing the path suffix to "scoring" in App.vue, resulting in the loss of the desired site UUID.


BryonLewis:
are you missing a commited file:  `src/actions/dataSource` that contains the function `reloadDataSourceFromStorage`?

BryonLewis:
We may want to have a discussion about the removal of a feature:
Before this the URL could also specify the current mode.  I.E if you went to `http://localhost:3000/#/KR_R002/` you would be using the RGD database but if you went directly to `http://localhost:3000/#/scoring/KR_R002` it would be in the scoring DB.  This way users could share a direct URL independent of having their local setting configured for a specific DB.
I don't know if this is useful so we may be able to remove the feature.

floryst:
There also appears to be some priority-based behavior, and it would be good to review them again to make sure they align with our requirements.
- visiting `/#/KR_R002` with no local storage key `databaseSource` will use the RGD DB
- visiting `/#/KR_R002` when local storage key `databaseSource` is `Scoring` will redirect to the scoring DB
- visiting `/#/scoring/KR_R002` will use the scoring DB regardless of the `databaseSource` value

BryonLewis:
> * visiting `/#/KR_R002` with no local storage key `databaseSource` will use the RGD DB
> * visiting `/#/KR_R002` when local storage key `databaseSource` is `Scoring` will redirect to the scoring DB
> * visiting `/#/scoring/KR_R002` will use the scoring DB regardless of the `databaseSource` value

I think that is correct, we want to default to RGD DB if no key is set. 
If you have configured it to default to a ScoringDB you want it to go there by default
Directly visiting a /scoring endpoint should take you to the scoring DB independent of a configured key.

BryonLewis (approved):
:+1:

PR #520, opened 2024-10-16
Optimize model run listing

zachmullen:
- [x] Remove unnecessary aliases
- [x] Remove ground truth aggregate computation and rely on ModelRun's ground_truth field instead
- [x] Short-circuit matching Region lookup (not in hot path)
- [x] Remove caching infrastructure since this branch should obviate it
- [x] Add denormalized fields to ModelRun for caching expensive summary statistics
- [x] Add method to compute the new fields on a single ModelRun
- [x] Add function to compute the new fields on all ModelRuns (mostly for use during migration)
- [x] Call computation function in all places where data might change


zachmullen:
As of 8d72baf, total response time of all queries for the model run listing endpoint is down to tens of ms:

Screenshot 2024-10-17 at 1 17 04 PM

Quite a few further optimizations are possible.

* We only need to compute the aggregate bounding box and time range when filters change, not when a new page of results is requested.
* We could also cache the properties of model runs that require joins/subqueries against other model runs, but that would likely require much more complicated invalidation logic. If the cardinality of model runs stays in the hundreds, this isn't worth doing.
* We could index certain fields of model runs, but again that won't be necessary if the number of model runs is on the order of hundreds.
* We could switch to cursor-based pagination, but that would have negligible impact until there are far more model runs in the database.

The thing remaining to do on this issue is to add the calls to the stats computation method in any of the places we need to.

rdwatch/core/db/functions.py
```diff
@@ -78,6 +78,12 @@ def __init__(self, field):
         return super().__init__(json_str, JSONField())  # noqa: B037
 
 
+class AsGeoJSONDeserialized(Cast):
```
zachmullen, line 81: If there's an easier way of getting `AsGeoJSON`'s result into a deserialized python object, I'd love to know it.

zachmullen:
This is ready for review. It contains a breaking change for model runs that are created through the REST API rather than processed via the ModelRunUpload pathway. Namely, once all data underneath a model run has been added, it is the clients' responsibility to POST to the new `/api/model-runs/{id}/finalization/` endpoint, which will ensure all the relevant stats are computed and saved on the model run.

mvandenburgh:
> This is ready for review. It contains a breaking change for model runs that are created through the REST API rather than processed via the ModelRunUpload pathway. Namely, once all data underneath a model run has been added, it is the clients' responsibility to POST to the new `/api/model-runs/{id}/finalization/` endpoint, which will ensure all the relevant stats are computed and saved on the model run.

Sorry, I missed this message @zachmullen. I'll take a look at this

rdwatch/core/views/model_run.py
```diff
@@ -306,45 +289,31 @@ def paginate_queryset(
...
+        model_runs = qs | queryset.aggregate(**aggregate_kwargs)
+        if filters.region:
+            if qs['count'] == 0:  # No model runs; we set bbox to Region bbox
+                region_filter = filters.dict()['region']
+                # We do a full table scan because this isn't a common case and isn't a bottleneck;
+                # the equivalent database query is really annoying due to the logic of the
+                # `Region.value` property.
+                for region in Region.objects.all():
+                    if region.value == region_filter:
+                        model_runs['bbox'] = region.geojson
```
mvandenburgh, line 314: I think we _might_ be able to replace the `Region.value` property with a `GeneratedField` to get around this. But, I think that's fine to postpone for the future

rdwatch/core/models/model_run.py
```diff
@@ -54,5 +57,96 @@ class ProposalStatus(models.TextChoices):
...
+        if recompute_all:
+            qs = cls.objects.all()
+        else:
+            qs = cls.objects.filter(cached_numsites=None)
+
+        # Unfortunately, Django ORM doesn't support joins in update queries, so doing this in a
+        # single join query would require raw SQL that would be a pain to maintain.
+        # Since this should only rarely run, we don't mind looping here.
+        for model_run in qs:
+            model_run.compute_aggregate_stats()
```
mvandenburgh, line 152: The number of model runs in the system could be significant. I believe the T&E deployment has tens of thousands of them. I'm not sure if it's better to use `.iterator()` here to avoid loading all that data into memory at once (at the cost of speed), or if that scale of data is still okay to do a regular iteration over, any thoughts?

mvandenburgh (changes requested):
These scripts will also need to be updated to use the new finalization endpoint -
https://github.com/ResonantGeoData/RD-WATCH/blob/optimize-model-run-list-endpoint/scripts/loadGroundTruth.py
https://github.com/ResonantGeoData/RD-WATCH/blob/optimize-model-run-list-endpoint/scripts/loadModelRun.py

rdwatch/scoring/views/site_image.py
```diff
@@ -149,9 +153,11 @@ def site_images(request: HttpRequest, id: UUID4):
...
                     )
                 ),
-                notes=F(site_db_model_cols['notes'])
-                if site_db_model_cols['notes']
-                else Value(''),  # TODO
+                notes=(
+                    F(site_db_model_cols['notes'])
+                    if site_db_model_cols['notes']
+                    else Value('')
+                ),  # TODO
```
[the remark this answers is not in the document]
zachmullen: Yeah, this was just `pre-commit` doing its thing. Not sure why this file was touched...

rdwatch/core/views/model_run.py
zachmullen: Yeah, this logic existed prior to my change and isn't really related, but I wanted to add a comment about it since it looks suspicious.

zachmullen:
@mvandenburgh what were you doing when you ran that test? Is there a stack trace associated?

mvandenburgh:
> @mvandenburgh what were you doing when you ran that test? Is there a stack trace associated?

Sorry, that lacked context - here's the full stack trace (happens on the `/api/model-runs/` endpoint)

```
more than one row returned by a subquery used as an expression
  File "/home/local/KHQ/michael.vandenburgh/.cache/pypoetry/virtualenvs/rdwatch-r8XUgLGN-py3.11/lib/python3.11/site-packages/django/db/backends/utils.py", line 105, in _execute
    return self.cursor.execute(sql, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
psycopg2.errors.CardinalityViolation: more than one row returned by a subquery used as an expression


The above exception was the direct cause of the following exception:

  File "/home/local/KHQ/michael.vandenburgh/.cache/pypoetry/virtualenvs/rdwatch-r8XUgLGN-py3.11/lib/python3.11/site-packages/ninja/operation.py", line 104, in run
    result = self.view_func(request, **values)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/local/KHQ/michael.vandenburgh/.cache/pypoetry/virtualenvs/rdwatch-r8XUgLGN-py3.11/lib/python3.11/site-packages/ninja/pagination.py", line 146, in view_with_pagination
    result[paginator.items_attribute] = list(result[paginator.items_attribute])
                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/local/KHQ/michael.vandenburgh/.cache/pypoetry/virtualenvs/rdwatch-r8XUgLGN-py3.11/lib/python3.11/site-packages/django/db/models/query.py", line 400, in __iter__
    self._fetch_all()
  File "/home/local/KHQ/michael.vandenburgh/.cache/pypoetry/virtualenvs/rdwatch-r8XUgLGN-py3.11/lib/python3.11/site-packages/django/db/models/query.py", line 1928, in _fetch_all
    self._result_cache = list(self._iterable_class(self))
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/local/KHQ/michael.vandenburgh/.cache/pypoetry/virtualenvs/rdwatch-r8XUgLGN-py3.11/lib/python3.11/site-packages/django/db/models/query.py", line 91, in __iter__
    results = compiler.execute_sql(
              ^^^^^^^^^^^^^^^^^^^^^
  File "/home/local/KHQ/michael.vandenburgh/.cache/pypoetry/virtualenvs/rdwatch-r8XUgLGN-py3.11/lib/python3.11/site-packages/django/db/models/sql/compiler.py", line 1562, in execute_sql
    cursor.execute(sql, params)
  File "/home/local/KHQ/michael.vandenburgh/.cache/pypoetry/virtualenvs/rdwatch-r8XUgLGN-py3.11/lib/python3.11/site-packages/debug_toolbar/panels/sql/tracking.py", line 235, in execute
    return self._record(super().execute, sql, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/local/KHQ/michael.vandenburgh/.cache/pypoetry/virtualenvs/rdwatch-r8XUgLGN-py3.11/lib/python3.11/site-packages/debug_toolbar/panels/sql/tracking.py", line 160, in _record
    return method(sql, params)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/local/KHQ/michael.vandenburgh/.cache/pypoetry/virtualenvs/rdwatch-r8XUgLGN-py3.11/lib/python3.11/site-packages/django/db/backends/utils.py", line 122, in execute
    return super().execute(sql, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/local/KHQ/michael.vandenburgh/.cache/pypoetry/virtualenvs/rdwatch-r8XUgLGN-py3.11/lib/python3.11/site-packages/django/db/backends/utils.py", line 79, in execute
    return self._execute_with_wrappers(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/local/KHQ/michael.vandenburgh/.cache/pypoetry/virtualenvs/rdwatch-r8XUgLGN-py3.11/lib/python3.11/site-packages/django/db/backends/utils.py", line 92, in _execute_with_wrappers
    return executor(sql, params, many, context)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/local/KHQ/michael.vandenburgh/.cache/pypoetry/virtualenvs/rdwatch-r8XUgLGN-py3.11/lib/python3.11/site-packages/django/db/backends/utils.py", line 100, in _execute
    with self.db.wrap_database_errors:
  File "/home/local/KHQ/michael.vandenburgh/.cache/pypoetry/virtualenvs/rdwatch-r8XUgLGN-py3.11/lib/python3.11/site-packages/django/db/utils.py", line 91, in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "/home/local/KHQ/michael.vandenburgh/.cache/pypoetry/virtualenvs/rdwatch-r8XUgLGN-py3.11/lib/python3.11/site-packages/django/db/backends/utils.py", line 105, in _execute
    return self.cursor.execute(sql, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
django.db.utils.ProgrammingError: more than one row returned by a subquery used as an expression

Internal Server Error: /api/model-runs/
[18/Nov/2024 10:37:30] "GET /api/model-runs/?groundtruth=false&page=1 HTTP/1.1" 500 4208
```

mvandenburgh:
Ok so it seems like this started happening after I ran the `loadGroundTruth` script. I dropped my database and reran it on `main`, and I am getting this error there as well. So it seems unrelated to this PR.

zachmullen:
Ok, makes sense. I was scratching my head trying to figure out how my changes could have caused that 😅

rdwatch/core/views/model_run.py
```diff
@@ -388,30 +357,30 @@ def create_model_run(
...
+    '/{id}/finalization/',
+    response={200: ModelRunDetailSchema},
+    auth=[ModelRunAuth()],
+)
+# this is safe because we're using a nonstandard header w/ API Key for auth
+@csrf_exempt
+def finalize_model_run(request: HttpRequest, id: UUID4):
+    model_run = get_object_or_404(ModelRun, pk=id)
+    model_run.compute_aggregate_stats()
+    return model_run
```
mvandenburgh, line 381: `ModelRunDetailSchema` includes some additional fields not on `ModelRun`. When we use it in the `create_model_run` endpoint above, we explicitly return a dictionary that is the expected structure, like this -
```suggestion
@router.post(
    '/{id}/finalization/',
    response={200: ModelRunDetailSchema},
    auth=[ModelRunAuth()],
)
# this is safe because we're using a nonstandard header w/ API Key for auth
@csrf_exempt
def finalize_model_run(request: HttpRequest, id: UUID4):
    model_run = get_object_or_404(ModelRun, pk=id)
    model_run.compute_aggregate_stats()
    return {
        'id': model_run.pk,
        'title': model_run.title,
        'performer': {
            'id': model_run.performer.pk,
            'team_name': model_run.performer.team_name,
            'short_code': model_run.performer.short_code,
        },
        'region_name': model_run.region.name,
        'parameters': model_run.parameters,
        'numsites': 0,
        'created': model_run.created,
        'expiration_time': model_run.expiration_time,
    }
```

Currently, the `/finalize/` endpoint always returns a 500 error because of this.

mvandenburgh (changes requested):
Looks good except for one issue with the `finalize` endpoint. After applying the fix I suggested locally, this functions as intended :+1:

rdwatch/core/views/model_run.py
zachmullen: Ah good catch, thanks

mvandenburgh (approved):

PR #521, opened 2024-10-17
Fix site image URIs

floryst:
SiteImage objects no longer have a `aws_location` property. Instead, they now have a `uri_locations` property, which is a list of URIs pointing to the assets that make up the SiteImage. Currently, S2 imagery is comprised of a single asset URI, while L8 imagery is comprised of 3 asset URIs (red, green, and blue).


BryonLewis (approved):
Thanks for the fast turnaround on this.

I pulled and happen to actually test the S2/L8 this time (I believe last time I only tested L8). Given only the NITF support for WV we can adjust that in the future once we have COG based NITF files.

I did have an issue with migrations of the scoring DB, but that could be that I had some borked rows from testing of point downloading.  Just noting it incase we hit any issues with migrating deployed versions.

Also thanks for moving the  `get_site_model_feature_JSON` function into the task.  I think initially I did dev of that in the request for testing of a single site before I brought it into a task to work for an entire modelrun.

floryst:
I pushed a new commit that should fix the scoring DB migrations. I wasn't saving the SiteImage to the correct database.

BryonLewis (approved):
Tested on the macbook that hasn't been migrated in a while and already had some scoring site images.  The migration worked properly and downloading images afterwards worked as well.

PR #522, opened 2024-10-17
Top Bar Redesign

BryonLewis:
Implement the new top bar redesign: https://www.figma.com/design/HzfMOKyPaKVVDN7OEKGfgs/RDWatch?node-id=912-3161&node-type=frame&t=N4IryMPZ3iEtguqt-0

- Update Theme to utilize colors from Figma

TODO:
- [x] Restructure top bar into a row with the legend also available in it,
- [x] Restructure legend into a button that togglesd into the legend with a close option
- [x] Observations/Sites are now buttons with a hover arrow on the right side.  The buttons change size based on the selection of Models or Ground Truth.
- [x] Region is also a button with a  dropdown that can have a disabled state as well.
- [x] Satellite images should probably still use the satellite icon to initiate downloading, it has a spinning icon for downloading and a dropdown for all the settings once it is configured.
    - [x] I think for satellite images to work properly I may need to add the current ModelRunList to the store so it can be accessed by other components instead of the current Event emitting that is being run right now.
- [x] Maybe remove the autoZoom functionality?
- [x] All buttons in the area need a tooltip to indicate what they do
- [x] Needs to handle squashing if the legend becomes wide (it has both observations and Site labels displayed).  Looks like I use some rows with columns that can squash when needed: 
- [x] Annotator Mode Toolbar needs to be modified with Add proposals
- [x] Buttons in the SideBar need to be removed that duplicated in the top bar.
image
image
image
image


vue/src/components/topbar/AddProposal.vue
floryst, line 1: Still need this file? It's imported elsewhere, but the file is empty.

BryonLewis: It's a topbar component so I was thinking about adding it under ./topbar but it's also a proposal component.  I'll leave it under the ./annotationViewer folder.

PR #524, opened 2024-10-22
Allow for viewing local geojson layers

floryst:
Users can now visualize local geojson files in RD-WATCH.

[Screen Recording 2024-10-22 at 3.19.43 PM.webm](https://github.com/user-attachments/assets/292e3f1a-be57-4b3b-890c-b89e194b667c)

Nice-to-haves and other considerations:
* toggle layer visibility
* nicer layer names (based on what?)
* edit geometry styles
* should the annotator view also support viewing local geojsons?

Other changes:
* The map is no longer partially covered by the sidebar
* An event is used for triggering `map.fitBounds()` instead of a state variable
* minor whitespace changes that were automatically done by my editor

### DIFF
--- a/vue/src/components/LocalLayers.vue
+++ b/vue/src/components/LocalLayers.vue
@@ -1,13 +1,15 @@
 <script setup lang="ts">
-import { computed } from 'vue';
+import { computed, ref } from 'vue';
 import { addLocalMapLayer, removeLocalMapLayer } from '../actions/localMapLayers';
 import { state } from '../store';
 import { getGeoJSONBounds } from '../utils';
 import { FitBoundsEvent } from '../actions/map';
 
+const isLoading = ref(false);
+
 async function loadFiles(files: File[]) {
-  await Promise.all(files.map(async (file) => {
-    const data = await new Promise<string>((resolve, reject) => {
+  const promise = Promise.all(files.map(async (file) => {
+    return new Promise<string>((resolve, reject) => {
       const reader = new FileReader();
       reader.onload = function() {
         resolve(reader.result as string);
@@ -15,8 +17,18 @@ async function loadFiles(files: File[]) {
       reader.onerror = reject;
       reader.readAsText(file, 'utf-8');
     });
-    addLocalMapLayer(JSON.parse(data));
   }));
+
+  isLoading.value = true;
+  promise.finally(() => {
+    isLoading.value = false;
+  });
+
+  const geoJsons = await promise;
+
+  geoJsons.forEach((data) => {
+    addLocalMapLayer(JSON.parse(data));
+  });
 }
 
 async function promptForGeoJSONs() {
@@ -56,11 +68,13 @@ const localLayers = computed(() => {
 </script>
 
 <template>
-  <div>
+  <div class="h-100 overflow-y-auto">
     <div class="text-center">
       <v-btn
         variant="flat"
         color="secondary"
+        :disabled="isLoading"
+        :loading="isLoading"
         @click="promptForGeoJSONs"
       >
         <template #prepend>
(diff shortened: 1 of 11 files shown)


vue/src/components/MapLibre.vue
```diff
@@ -151,50 +152,73 @@ watch([() => state.timestamp, () => state.filters, () => state.satellite, () =>
...
+
+  const addedIds = newIds.filter((id) => !loadedLocalLayerIds.includes(id));
+  const removedIds = loadedLocalLayerIds.filter((id) => !newIds.includes(id));
+
+  loadedLocalLayerIds = [...newIds];
+
+  removedIds.forEach((id) => {
+    map.value?.removeLayer(`local-layer-${id}:fill`);
+    map.value?.removeLayer(`local-layer-${id}:outline`);
+    map.value?.removeSource(`local-source-${id}`);
```
BryonLewis, line 168: This was something done before my time with the project, but toggling different modelRuns basically rebuilds the entire data passed to the model.  That means that switching tabs to the model runs and turning on any model run will make the local geoJSON features disappear.  It may be necessary to add some data to the the sources in the main style function:
https://github.com/ResonantGeoData/RD-WATCH/blob/main/vue/src/mapstyle/index.ts#L34

Not the way I would do stuff if starting from the beginning (I'd rather use maplibre's addSource/addLayer) but you probably need to add an additional function here to make sure the localGeoJSONs don't get disabled when running.

vue/src/components/LocalLayers.vue
```diff
@@ -0,0 +1,131 @@
...
+        No local layers
+      </div>
+    </div>
+    <v-list>
+      <v-list-item
+        v-for="layer in localLayers"
+        :key="layer.id"
+        :title="`Layer ${layer.id}`"
+      >
+        <template #append>
```
BryonLewis, line 98: Besides focusing and deleting it may be nice to toggle layer visibility on/off.

vue/src/components/LocalLayers.vue
```diff
@@ -0,0 +1,131 @@
...
+        class="mt-8 text-grey-darken-2"
+      >
+        No local layers
+      </div>
+    </div>
+    <v-list>
+      <v-list-item
+        v-for="layer in localLayers"
+        :key="layer.id"
+        :title="`Layer ${layer.id}`"
```
BryonLewis, line 96: I think it's fine for a Layer to begin with a Static Id like this when initialized, it may be nice to have an edit field that way a user can upload multiple layers and give them some reasonable name.  Just an edit button and allowing the user to rename locally for organizational purposes.

BryonLewis (changes requested):
Sorry about the delay in reviewing this, just a couple of suggestions.

As usual some future  features not needed for this PR (mostly just thoughts):
- May be nice to integrate with tooltips and present a table of the properties for a feature, or perhaps selection and provides a table of the properties for the selected features.  There could be metadata associated with their geometry that help distinguish items.

vue/src/components/LocalLayers.vue
floryst: I opted to just use an editable box in place of the static name.

vue/src/components/LocalLayers.vue
floryst: Added layer visibility toggles

vue/src/components/MapLibre.vue
floryst: I fixed the issue and moved the layer drawing into the `style` function.

BryonLewis (approved):
Minor thing if you want to add it.  I notice that on initial layer adding it focuses to the layer. Any subsequent layers being add it doesn't focus on the new layer.  Maybe this should be the case?

Thanks for making the other changes that I requested.

PR #529, opened 2024-10-30
NITF Pansharpening

BryonLewis:
- Limits downloading to 'nitf:compressing' == 'NC' or non-compressed NITF files.  C8 files are skipped
- Classifies images as panchromatic (having instruments = ['panchromatic'] or visual (having instruments = ['multi-vis']).  After getting the captures it will attempt to add `panuri` fields to any image that it can find that is panchromatic and within 60 seconds timestamp of the original visual image
- The replacing of the 'microsecond' for the time is because some images in the STAC server report with microseconds and others don't.  This unifies it so it is easier to find the same timestamp.
- The Try/Except for the animation export was turned off (probably for testing or checking errors) so I turned it back on.

TODO:
- [x] Testing for what the system does if it finds a compression other than 'NC' or 'C8' (error out?)
- [x] Handle other instruments like ['swir-multi']
    - Going to ignore anything other than ['panchromatic'] or ['multi-vis']
- [x] Getting an error 'WarpedVRT does not permit boundless reads'
- [x] Error: 'bounds and transform are inconsistent'
    - Not entirely sure but this seems like an issue where the Metadata for a NITF doesn't match the NITF itself.  If I force riotiler to ignore the error it creates an upside-down image.  

NOTES:

- The 'bounds and transform are inconsistent' and the 'WarpedVRT does not permit boundless reads' are both currently being caught and logged as 'info' in the system to allow for the downloading of a  majority of images.


rdwatch/core/utils/worldview_nitf/raster_tile.py
```diff
@@ -9,12 +9,39 @@
...
+    elif gray > 0:
+        return gray
+    else:
+        raise ValueError(
+            f'colorInterp: {colorinterp} for files does not have RGB or Gray channels'
+        )
+
+
 @contextmanager
 def _download_nitf_image(uri: str) -> Generator[str, None, None]:
```
mvandenburgh, line 46: Can this function be safely removed?

rdwatch/core/utils/worldview_nitf/raster_tile.py
```diff
@@ -61,25 +90,87 @@ def get_worldview_nitf_bbox(
...
-        return rgb.render(img_format=format)
+        logger.info(f'Downloading WorldView NITF bbox: {bbox} scale: {scale}')
+        if capture.panuri is not None:
+            with Reader(input=capture.panuri) as pan_img:
+                logger.info(f'Downloading PanURI Chip: {capture.panuri}')
+                try:
+                    pan_chip = pan_img.part(bbox=bbox, dst_crs='epsg:4326')
+                except ValueError as e:
+                    logger.info(f'Value Error: {e} - {capture.panuri}')
+                    return None
```
mvandenburgh, line 105: What circumstances would cause a `ValueError` to be raised? Should we be letting it get thrown so it gets recorded in Sentry, or are we okay with just silently suppressing it.

rdwatch/scoring/tasks/animation_export.py
```diff
@@ -601,24 +601,24 @@ def create_site_animation_export(
...
+        site_export.name = name
+        with open(file_path, 'rb') as file:
+            site_export.export_file.save(name, File(file))
+        site_export.save()
+        if os.path.exists(file_path):
+            os.remove(file_path)
+    except Exception as e:
+        logger.warning(f'Error when processing Animation: {e}')
+        if site_export:
+            site_export.delete()
```
mvandenburgh, line 621: Same question here. If an animation fails to generate, would it be better to get an exception in Sentry instead of suppressing it?
```suggestion
    try:
        site_export.celery_id = task_id
        site_export.save()
        file_path, name = create_animation(self, site_evaluation_id, settings)
        if file_path is False and name is False:
            logger.warning('No Images were found')
            site_export.delete()
            return
        site_export.name = name
        with open(file_path, 'rb') as file:
            site_export.export_file.save(name, File(file))
        site_export.save()
        if os.path.exists(file_path):
            os.remove(file_path)
    finally:
        logger.warning(f'Error when processing Animation: {e}')
        if site_export:
            site_export.delete()
```

rdwatch/core/utils/worldview_nitf/raster_tile.py
```diff
@@ -61,25 +90,87 @@ def get_worldview_nitf_bbox(
...
+                            vis_chip = vis_img.part(
+                                bbox=bbox,
+                                indexes=rgb_channels,
+                                dst_crs='epsg:4326',
+                                width=pan_chip.width,
+                                height=pan_chip.height,
+                            )
+                        except ValueError as e:
+                            logger.info(f'Value Error: {e} - {capture.uri}')
+                            return None
```
mvandenburgh, line 123: Same question

rdwatch/core/utils/worldview_nitf/raster_tile.py
```diff
@@ -61,25 +90,87 @@ def get_worldview_nitf_bbox(
...
+                ):  # Ensure all RGB channels exist
+                    try:
+                        final_chip = vis_img.part(
+                            bbox=bbox,
+                            dst_crs='epsg:4326',
+                            indexes=rgb_channels,
+                        )
+                    except ValueError as e:
+                        logger.info(f'Value Error: {e} - {capture.uri}')
+                        return None
```
mvandenburgh, line 162: Same question

rdwatch/core/utils/worldview_nitf/raster_tile.py
BryonLewis: yeah I think we can unless if we're going to support any compressed files in the future.

rdwatch/core/utils/worldview_nitf/raster_tile.py
BryonLewis: I can't be sure because I don't know the entire context of all the files in the STAC collection or a guarantee of the similarity of the files.  If I allow errors to propagate up we then have a bunch of failed downloaded images.  I.E. that image could be at the beginning of a list of site downloads, breaking all images from downloading if you're downloading all sites.  I'm open to a way of recording this in sentry without causing the tasks to exit.  Most of this PR has been spent testing and finding new things about these images that completely break downloading images.  I'm sure more of these corner cases exist but I don't have the time to find all of them.  This allows the files that work to be downloaded while skipping the files don't without killing the entire task the first instance it hits a weird image in the list of 200 images.

rdwatch/core/utils/worldview_nitf/raster_tile.py
```diff
@@ -61,25 +90,87 @@ def get_worldview_nitf_bbox(
...
-
-        return rgb.render(img_format=format)
+        logger.info(f'Downloading WorldView NITF bbox: {bbox} scale: {scale}')
+        if capture.panuri is not None:
+            with Reader(input=capture.panuri) as pan_img:
+                logger.info(f'Downloading PanURI Chip: {capture.panuri}')
+                try:
+                    pan_chip = pan_img.part(bbox=bbox, dst_crs='epsg:4326')
+                except ValueError as e:
+                    logger.info(f'Value Error: {e} - {capture.panuri}')
```
mvandenburgh, line 104:
```suggestion
                    logger.info(f'Value Error: {e} - {capture.panuri}')
                    sentry_sdk.capture_exception(e)
```

rdwatch/core/utils/worldview_nitf/raster_tile.py
mvandenburgh, line 123:
```suggestion
                            logger.info(f'Value Error: {e} - {capture.uri}')
                            sentry_sdk.capture_exception(e)
                            return None
```

rdwatch/core/utils/worldview_nitf/raster_tile.py
```diff
@@ -61,25 +90,87 @@ def get_worldview_nitf_bbox(
...
+                    channel is not None for channel in rgb_channels
+                ):  # Ensure all RGB channels exist
+                    try:
+                        final_chip = vis_img.part(
+                            bbox=bbox,
+                            dst_crs='epsg:4326',
+                            indexes=rgb_channels,
+                        )
+                    except ValueError as e:
+                        logger.info(f'Value Error: {e} - {capture.uri}')
```
mvandenburgh, line 161:
```suggestion
                        logger.info(f'Value Error: {e} - {capture.uri}')
                        sentry_sdk.capture_exception(e)
```

rdwatch/scoring/tasks/animation_export.py
mvandenburgh, line 621:
```suggestion
    except Exception as e:
        sentry_sdk.capture_exception(e)
        if site_export:
            site_export.delete()
```

mvandenburgh (changes requested):
These changes will propagate any unexpected exceptions to Sentry, without interrupting execution.

rdwatch/scoring/tasks/animation_export.py
mvandenburgh: > I'm open to some way to log this to sentry without causing the task to exit.

Yes, this is possible - see https://github.com/ResonantGeoData/RD-WATCH/pull/529#pullrequestreview-2441268767

mvandenburgh (approved):
Tested and everything seems to work as expected in both RGD and Scoring mode :+1:

floryst (approved):

PR #531, opened 2024-11-05
Initial IQR integration

floryst:
Adds IQR support


BryonLewis:
Generic Update Notes:

- P/N buttons to mdi-check + mdi-close buttons
- Possible indication of loading for refining and the initial seed site being loaded in (progress indeterminate or something)
- When clicking on a site it should zoom in, turn on the vector tiles for that model-run, turn on the image and possible bring up the image viewer as well.  Currently it just zooms to the bounding box for that site without displaying vector tiles. It's a bit weird because we currently have modelRuns tried to their region but this should be possible in the front-end code (it may turn off the current region to swap to the new one)

floryst:
Updates:
- IQR sidebar now more closely matches the mockups. UX is improved with icons, loading indicators, and colors.
- Initial attempt at generating vector tiles for the IQR results. There is some performance issue going on with the tiles rendering late, likely due to waterfall effects with network requests based on initial observations.

image

floryst:
Updates:
- updated with main branch
- fixed IQR results not returning 50 results, when the underlying result list is > 50
- selecting an IQR result will show the image viewer, and the map image will reflect the current image viewer image.
- closing the IQR panel will clear the current IQR session
- IQR result polygons now show on top

BryonLewis (approved):
Just leaving some future notes for this PR:

- Eventually updating the system so that the polygons/annotations that are associated with a site that is selected through IQR corresponds to a similar vector tile for the region.  I.E filtering or turning of the vector tile from IQR if the vector tile from a model run with the same site uuid is matching.
- Update the Image preview to provide image advancement on hover.  I.E having the images go through their standard list on hover.
- Some better resizing of the images to maintain some consistency between the items in the IQR list.

floryst:
I'll move your comments to an issue so it can be picked up later.

PR #533, opened 2024-11-07
Image viewer line thickness

floryst:
A new UI element, "line thickness factor", provides users with the means to adjust the polygon line thickness in both the viewer and the animation export.

Other changes:
- show the point size slider in the animation dialog
- sync the point size + rescale parameters to the animation dialog

New line width controls:

image

### DIFF
--- a/rdwatch/core/tasks/animation_export.py
+++ b/rdwatch/core/tasks/animation_export.py
@@ -304,6 +304,7 @@ class GenerateAnimationSchema(BaseModel):
     include: list[Literal['obs', 'nonobs']] = ['obs', 'nonobs']
     rescale: bool = False
     rescale_border: int = Field(1, ge=0)  # Border rescale should be an integer >= 0
+    line_thickness_factor: float = 1.0
 
 
 @shared_task
@@ -321,6 +322,7 @@ def create_animation(self, site_evaluation_id: UUID4, settings: dict[str, Any]):
         'noData': settingsSchema.noData,
         'include': settingsSchema.include,
     }
+    line_thickness_factor = settingsSchema.line_thickness_factor
 
     # Fetch the SiteEvaluation instance
     try:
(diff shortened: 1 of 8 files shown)


BryonLewis:
This is only changing the client side visualization for the line (which is good).  This option also needs to be added to the scoring/rgd app tasks for exporting Animations as well. 
https://github.com/ResonantGeoData/RD-WATCH/blob/main/rdwatch/core/tasks/animation_export.py#L310
https://github.com/ResonantGeoData/RD-WATCH/blob/main/rdwatch/scoring/tasks/animation_export.py#L140

This may have some relevance to: #534 where it will pass the new line-thickness setting to the exporting function.

floryst:
@BryonLewis this is ready for a re-review.

rdwatch/core/tasks/animation_export.py
```diff
@@ -540,7 +542,10 @@ def create_animation(self, site_evaluation_id: UUID4, settings: dict[str, Any]):
                     for lon, lat in transformed_coords
                 ]
                 color = label_mapped.get('color', (255, 255, 255))
-                draw.polygon(pixel_coords, outline=color)
+                polyline_width = int(
+                    (max(max_height_px, max_width_px) * line_thickness_factor) // 100
```
BryonLewis, line 546: you're int casting something that has already been floored by the `//` should this just be a single `/` to be closer to the Viewer version on the client.  It looked like in testing changing this would affect the value slightly (rounding up sometimes instead of down all the time).

rdwatch/core/tasks/animation_export.py
```diff
@@ -304,6 +304,7 @@ class GenerateAnimationSchema(BaseModel):
     include: list[Literal['obs', 'nonobs']] = ['obs', 'nonobs']
     rescale: bool = False
     rescale_border: int = Field(1, ge=0)  # Border rescale should be an integer >= 0
+    line_thickness_factor: float = 1.0
```
BryonLewis, line 307: This same stuff needs to be added to the ./rdwatch/scoring/tasks/animation_export.py to maintain parity.

BryonLewis (changes requested):
Looks great so far.  Just a minor thing regarding the line thickness with the export functionality as well as adding the same logic to the scoring application.

I wouldn't worry about it now, but if line thickness becomes too high you can get disjointed corners.  I don't think it's within the scope of this PR but we could modify the polygon to draw lines and specify the type of joint between them instead of using ImageDraw's polygon function that doesn't have that. 
image

rdwatch/core/tasks/animation_export.py
floryst: Resolved. I decided do floor(X / 100).

rdwatch/core/tasks/animation_export.py
floryst: Added to `scoring/animation_export.py`.

BryonLewis (approved):

PR #535, opened 2024-11-13
Pin postgis docker version

floryst:
Existing deployment databases (postgresql v16) are not compatible with postgresql v17. Pin to v16 so we don't become incompatible with existing databases if the docker image gets deleted and re-downloaded.


mvandenburgh (commented):
I think we should also pin the `scoringdb` service.

mvandenburgh (approved):

PR #537, opened 2024-11-14
Optimize sites list query

mvandenburgh:
Moves the color_code subquery to a separate query to avoid a slow subquery in the main query, due to the
EvaluationBroadAreaSearchDetection table being extremely large. The number of sites in an evaluation run is small enough that it's feasible to load the site_id->color_code mapping into memory and manually join it with the sites_list in Python.

### DIFF
--- a/rdwatch/scoring/views/model_run.py
+++ b/rdwatch/scoring/views/model_run.py
@@ -697,6 +697,7 @@ def get_sites_query(model_run_id: UUID4):
         sites=JSONBAgg(
             JSONObject(
                 id='pk',
+                site_id='site_id',
                 number=Substr(F('site_id'), 9, 4),  # pos is 1 indexed,
                 bbox=BoundingBox(
                     Func(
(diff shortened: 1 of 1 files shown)


zachmullen (approved):

BryonLewis (approved):

rdwatch/scoring/views/model_run.py
```diff
@@ -697,6 +697,7 @@ def get_sites_query(model_run_id: UUID4):
         sites=JSONBAgg(
             JSONObject(
                 id='pk',
+                site_id='site_id',
```
zachmullen, line 700: Is this setting this to the literal string `'site_id'` or is it automatically doing `F('site_id')`?

mvandenburgh: It's automatically doing `F('site_id')`. I thought the same thing when I saw `id='pk'`.

PR #538, opened 2024-11-14
Show site observation date in export

floryst:
If a site image has an associated site observation, show the site observation's date. Otherwise, show "-obs".

A sample screenshot is provided below. (The observation date is arbitrary and not indicative of real data.)

image

### DIFF
--- a/rdwatch/core/tasks/animation_export.py
+++ b/rdwatch/core/tasks/animation_export.py
@@ -446,6 +446,7 @@ def create_animation(self, site_evaluation_id: UUID4, settings: dict[str, Any]):
     frames = []
     np_array = []
     polygon = None
+    last_obs_date = None
     point = None
     count = 0
     for img, width, height, image_record in images_data:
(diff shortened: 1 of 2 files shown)


floryst:
So in the `-obs` case the `observation` object is None, so which date should we use?

BryonLewis:
> So in the `-obs` case the `observation` object is None, so which date should we use?

The previously stored obs value.  It should be storing some obs value because that's how it knows the label and the polygon.

floryst:
Updates:
- Show the "last observation date". @BryonLewis do we want to show the initial last observation date as empty (e.g. "-----"), or as the first image date?

image

BryonLewis (approved):
Tested the newer updates and it works.  We may in the future might want to make it clearer when the dates don't match but I don't know a great way to do that now.

BryonLewis:
> How about coloring the box to indicate such a state? Could be green for matching, grey for not matching. Or red for not matching, grey for matching. It would depend on if we want to emphasize matching vs not matching dates.

I would just leave it normal if the match and maybe a slightly different color if they are different.  Not like a bright red, maybe like a light/pastel orange/yellow.  I would just be worried if you set the FPS to like 5 and have the red/green flashing it could be too intensive.

PR #541, opened 2024-11-18
Update to Python 3.13

mvandenburgh:
Updates the backend to Python 3.13 (currently we are on 3.11).


mvandenburgh:
Not sure what's going on here, but poetry struggles to install dependencies on both Python 3.12 and 3.13. Marking this as draft and will come back to this after the rest of the dependency upgrades are finished.

mvandenburgh:
I won't be working on this PR for the foreseeable future.

PR #546, opened 2024-11-19
Fix scrolling model lists and local layers

floryst:
I noticed that scrolling long model lists and local layers is broken. Updated CSS to properly scroll on those long lists.

The changes in `SideBar.vue` look extensive, but in reality I only modified some classes and nested all of the elements from `<mode-selector />` to `<EvalFilter />` in a `<div class="flex-grow-0">`.

### DIFF
--- a/vue/src/components/LocalLayers.vue
+++ b/vue/src/components/LocalLayers.vue
@@ -58,8 +58,8 @@ const localLayers = computed(() => {
 </script>
 
 <template>
-  <div class="h-100 overflow-y-auto">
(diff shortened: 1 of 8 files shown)


BryonLewis (commented):
Could you duplicate this for the Annotator view (./views/AnnotationViewer.vue, ./components/annotationViewer/SideBar.vue) to make sure that scrolls as well.

floryst:
I've fixed the scrolling for the annotation view.

BryonLewis (approved):

PR #552, opened 2024-11-21
fixes display of image-viewer in annotator mode

BryonLewis:
When testing the SAM stuff I noticed the image viewer wasn't coming up.  This is related to #543.  Added code so that in Annotator Mode the image viewer will display when selecting a site.

### DIFF
--- a/vue/src/views/AnnotationViewer.vue
+++ b/vue/src/views/AnnotationViewer.vue
@@ -82,7 +82,7 @@ const updateSiteList = async () => {
       :site-evaluation-name="state.selectedImageSite.siteName"
       :date-range="state.selectedImageSite.dateRange"
       editable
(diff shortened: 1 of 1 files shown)


floryst (approved):
Seems like there are just several different places where we duplicate the styles and structure. Not the most ideal; maybe we'll get a chance to properly refactor in the future.

BryonLewis:
> Seems like there are just several different places where we duplicate the styles and structure. Not the most ideal; maybe we'll get a chance to properly refactor in the future.

This was a result of the iterative development with focusing on newer features without definitive use cases.  The annotator view was initially designed for annotators to review sites quickly and then more features were added to it that I don't think even the annotators use.  Even now I don't know if there is a good understanding of how the annotator view will be used in the future.

PR #553, opened 2024-11-21
Optimize scoring vector tiles

mvandenburgh:
This removes the expensive `color_code` subquery and adds a new endpoint that returns the color mapping by itself. The endpoint is cached, but even on cache misses it is much faster than the previous method.

I've updated the client to handle this new approach. This required reworking how the color_code MapLibre filter worked quite a bit.

### DIFF
--- a/rdwatch/scoring/views/vector_tile.py
+++ b/rdwatch/scoring/views/vector_tile.py
@@ -16,9 +16,7 @@ from django.db.models import (
     Func,
     IntegerField,
     Min,
-    OuterRef,
     Q,
-    Subquery,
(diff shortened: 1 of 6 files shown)


mvandenburgh:
Oh, I wasn't paying attention to CI. I'll push a fix for those

mvandenburgh:
@BryonLewis I added a fix for the color code simple/detailed filter. Can you try this out again?

BryonLewis (approved):
Pulled, cleared my cache and it looks 'more' like what I expect the detailed/simple views to look like.

PR #558, opened 2024-11-22
Remove color_code from /sites/ endpoint, update client

mvandenburgh:
I missed some of this in #553. Also, I realized we don't even need to embed the `color_code` in the `/sites/` endpoint response anymore, since we now have the color code mapping endpoint that gets stored in the global Vue state.


BryonLewis (approved):
Did some quick testing and they showed up and the ordering when selecting was correct as well.  I was also able to download sites by using a custom date range to expand the one day of the ground truth sites.

PR #559, opened 2024-11-25
Configure sentry logging integration for ERROR logs only

mvandenburgh:
Setting this to WARNING causes to many extraneous logs to be reported as issues in Sentry. `rio_tiler` in particular is very noisy, and it's difficult to calibrate things because we have no control over the source data its processing.


BryonLewis (approved):

zachmullen:
Some warnings are helpful to see, so I think a better solution would be to suppress this specific warning at the python level using `warnings.filterwarning('ignore', ...)` with some combination of `message` and `module`.

